PR #2, opened 2016-11-23
role epel: allow enable repo on CentOS

dahorak:
It also contains check to forbid this functionality on RHEL.

### DIFF
--- a/roles/epel/tasks/main.yml
+++ b/roles/epel/tasks/main.yml
@@ -3,6 +3,10 @@
 # https://fedoraproject.org/wiki/EPEL
 # https://getfedora.org/keys/
 
+- name: Check if EPEL is not permanently enabled on RHEL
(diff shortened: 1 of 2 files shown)


roles/epel/tasks/main.yml
```diff
@@ -18,11 +18,15 @@
     name='https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version  }}.noarch.rpm'
     state=present
 
+- name: Check if EPEL is not permanently enabled on RHEL
+  fail: msg="Do not permanently enable EPEL on RHEL!"
+  when: enabled and ansible_distribution == 'RedHat'
+
```
mbukatov: Could you move this block to the beginning of the role, so that it would fail immediately not leaving the epel repo enabled after the fail?

mbukatov:
Looks good. I tested that and it works as expected.

PR #3, opened 2016-11-23
QE Server setup tweaks for jenkins slave

mbukatov:
Moving user tasks from qe server playbook into new ansible role for jenkins slave playbook to reuse it.


mbukatov:
@dahorak Does this change match what we have agreed on today?

PR #4, opened 2016-11-28
rh-python35 on RHEL 7 machine

dahorak:
Package centos-release-scl is available only on CentOS machine, not applicable on RHEL.


mbukatov:
That's good point. Such check would not hurt.

PR #5, opened 2016-11-29
Enable/disable epel globaly (on CentOS)

dahorak:


mbukatov:
We have just did a little cleanup of Filip's branches, which have been combined to a single one and rebased on top of master. It should be good enough for you to drop your branches with epel changes.

PR #8, opened 2016-12-02
Update qe server setup for CI

mbukatov:
I have added installation of Jonathan's ansible gluster module and moved both url and versions of git repositories into variables, which would allow ah hoc tweaking if needed.

### DIFF
--- a/roles/qe-server-user/tasks/main.yml
+++ b/roles/qe-server-user/tasks/main.yml
@@ -17,6 +17,9 @@
    - name: 'usmqe-tests'
      repo: 'https://github.com/Tendrl/usmqe-tests.git'
      version: 'master'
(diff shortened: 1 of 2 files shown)


mbukatov:
@dahorak check if it makes sense to you, I have tried it and it seems to work

roles/qe-server-user/tasks/main.yml
```diff
@@ -13,13 +13,13 @@
   with_items:
    - name: 'usmqe-setup'
      repo: "{{ usmqe_setup_repo }}"
-     version: 'master'
+     version: "{{ usmqe_setup_version }}"
```
dahorak: `version` is slightly strange name (I would think more about tags than branches), but that's just my feeling.

PR #9, opened 2016-12-05
Fix ansible config of qe server

mbukatov:


mbukatov:
@dahorak added the reconfiguration you suggested, also I noticed I had a bug in the forks reconfiguration, so this got fixed as well

PR #10, opened 2016-12-07
role qe-server: wrap variable by quotes for backward compatibility

dahorak:
role qe-server: wrap variable by quotes for backward compatibility with ansible 1.9

### DIFF
--- a/roles/qe-server-user/tasks/main.yml
+++ b/roles/qe-server-user/tasks/main.yml
@@ -31,7 +31,7 @@
   lineinfile:
     dest={{ ansible_user_dir }}/.bashrc
     state=present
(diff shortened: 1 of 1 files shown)


roles/qe-server-user/tasks/main.yml
```diff
@@ -31,7 +31,7 @@
   lineinfile:
     dest={{ ansible_user_dir }}/.bashrc
     state=present
-    line={{ item }}
+    line="{{ item }}"
```
mbukatov: Oo, I just tested it and it still works with `ansible-2.2.0.0-3.fc24.noarch`.

PR #12, opened 2016-12-21
Role qe-server: disable host_key_checking in ansible.cfg

dahorak:


mbukatov:
I would rather do this via change in `~/.ssh/config` file of usmqe user, so that this change would change behavior of all ssh tools.

dahorak:
@mbukatov would you rather fix this just in `~/.ssh/config` (and do not change the `ansible.cfg`) or do you agree to have this configured in both ways?

mbukatov:
It think it would be better to drop the change in global `/etc/ansible/ansible.cfg` so that the change is done on a single place.

dahorak:
Fixed issue with missing `~/.ssh/config` file. Tested and it works as expected.

PR #13, opened 2016-12-21
Add qe_server_sshkey.yml playbook (used mainly in CentOS CI)

dahorak:
For deploying public ssh keys for usmqe user to qe_server and usmqe nodes.


dahorak:
description updated

PR #14, opened 2016-12-21
Configure firewall on test machines (workaround)

dahorak:
This is just workaround, there needs to be proper firewall configuration.

### DIFF
--- a/qe_gluster_firewall.yml
+++ b/qe_gluster_firewall.yml
@@ -1,0 +1,16 @@
+---
+# HACK
+# FIXME: There should be proper firewall configuration!
+
+- hosts: gluster:usm_server
+  remote_user: root
+  tasks:
+  - name: stop firewall
+    service:
+      name: firewalld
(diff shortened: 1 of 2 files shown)


qe_gluster_firewall.yml
```diff
@@ -0,0 +1,16 @@
...
+- hosts: gluster:usm_server
+  remote_user: root
+  tasks:
+  - name: stop firewall
+    service:
+      name: firewalld
+      state: stopped
+      enabled: no
+    ignore_errors: yes
+  - name: pause for 10 second
```
mbukatov: Could you add a comment with quick explanation of this task, eg. why is this needed?

qe_gluster_firewall.yml
```diff
@@ -0,0 +1,16 @@
+---
```
mbukatov: This file should be named something like `firewall.tendlr_gluster.something.yml` :)

PR #15, opened 2017-01-03
add tendrl-frontend role

fbalak:


roles/tendrl-frontend/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+  when: install_from == "source"
+
+- include: binary.yml
+  when: install_from == "packages"
+
+#
+# post installation configuration
+#
+
+# as documented in README file
```
mbukatov: This block of comments may be dropped.

roles/tendrl-frontend/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+
+- include: source.yml
+  when: install_from == "source"
+
+- include: binary.yml
+  when: install_from == "packages"
+
+#
+# post installation configuration
+#
```
mbukatov: Is there really no post installation configuration, or do you plan to add some later? (I'm just curious)

fbalak: There is currently no post installation configuration. Something may be added later.

PR #16, opened 2017-01-03
tendrl_gluster_peer_bricks playbook

dahorak:
- Move `gluster peer probe` from playbook to role `qe-gluster-peers` (and remove it from `tendrl_gluster.yml` playbook).
- Add `gluster_peers_bricks.yml` playbook
  - Create gluster trusted storage pool and bricks on all available devices on gluster nodes.
- Add `tendrl_gluster_peers_bricks.yml` playbook.
  - Install Tendrl server and tendrl-gluster nodes.
  - Configure gluster trusted storage pool and create bricks on all available devices on gluster nodes.


dahorak:
@mbukatov do you agree with this structure?
In CI installation I would use tendrl_gluster_peers_bricks.yml playbook, but it will be possible to just install tendrl and gluster (without any configuration on gluster site) or configure gluster trusted storage pool and bricks without Tendrl server.

gluster_peers_bricks.yml
```diff
@@ -0,0 +1,12 @@
+---
+#
+# Create gluster trusted storage pool and bricks
+# on all available devices on gluster nodes.
+#
+
+- hosts: gluster
+  remote_user: root
+  roles:
+  #- { role: qe-gluster-peers, gluster_nodes: "{{groups['gluster']}}"}
```
mbukatov: Based on the description, it should be possible to:

> configure gluster trusted storage pool and bricks without Tendrl server

But for this to work, we would need to add `gluster-centos` role there.

mbukatov: Checking on my usm5 cluster (4 storage servers), I see that running role `cluster-centos` role again (so that the status is ok) takes just 2.73s, so it should not be slowing the setup down.

dahorak:
@mbukatov Ok, running gluster-centos role twice shouldn't break anything, so I added it to the gluster_peers_bricks playbook.

PR #20, opened 2017-01-05
add tendrl-performance-monitoring role

fbalak:
add tendrl-performance-monitoring role, add openssl-devel as prerequisity

### DIFF
--- a/roles/tendrl-performance-monitoring/tasks/binary.yml
+++ b/roles/tendrl-performance-monitoring/tasks/binary.yml
@@ -3,3 +3,6 @@
 #  yum:
 #    name=tendrl-performance-monitoring
 #    state=latest
+
+- fail: msg="There are no packages yet, we can't start and enable it"
+  when: not ignore_incomplete_installation
--- a/roles/tendrl-performance-monitoring/tasks/binary.yml
+++ b/roles/tendrl-performance-monitoring/tasks/binary.yml
@@ -1,0 +1,5 @@
+# TODO Add when packages are ready
+#- name: Install tendrl-performance-monitoring
+#  yum:
+#    name=tendrl-performance-monitoring
+#    state=latest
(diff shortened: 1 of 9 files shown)


roles/tendrl-performance-monitoring/tasks/binary.yml
```diff
@@ -0,0 +1,5 @@
+# TODO Add when packages are ready
+#- name: Install tendrl-performance-monitoring
+#  yum:
+#    name=tendrl-performance-monitoring
+#    state=latest
```
mbukatov: I guess that this is commented out because there is no build for `tendrl-performance-monitoring` yet, but I would rather keep it there so that when someone runs it and the package is still not there, ansbile run would fail with a clear error message. When I run this role with packages (which is the default) now, it would say "ok" - which is kind of misleading.

roles/tendrl-performance-monitoring/tasks/source.yml
```diff
@@ -0,0 +1,49 @@
...
+      dest: 'performance_monitoring_logging.yaml'
+
+- name: Specify log_cfg_path in tendrl.conf
+  ini_file:
+    dest=/etc/tendrl/tendrl.conf
+    section=performance_monitoring
+    option=log_cfg_path
+    value="/etc/tendrl/performance_monitoring_logging.yaml"
+
+# TODO add after daemon file is available
```
mbukatov: If the systemd file is not yet present, add a line like this to make it clear:

```
- fail: msg="There is no systemd unit yet, we can't start and enable it"
  when: not ignore_incomplete_installation
```

As I note in https://usmqe-tests.readthedocs.io/en/latest/test_env_setup.html#details-on-installation-from-sources

tendrl_gluster.yml
```diff
@@ -21,6 +21,7 @@
     - { role: epel, epel_enabled: 1 } # based on https://github.com/Tendrl/node_agent/issues/75
     - tendrl-common
     - tendrl-node-agent
+    - tendrl-performance-monitoring
```
mbukatov: I would drop the role from this playbook for now.

PR #21, opened 2017-01-06
add tendrl-alerting role

fbalak:
add tendrl-alerting role

### DIFF
--- a/roles/tendrl-alerting/tasks/source.yml
+++ b/roles/tendrl-alerting/tasks/source.yml
@@ -1,0 +1,51 @@
+#
+# Installation
+#
+
+- name: Directory for git repository
+  file:
+    path="{{ tendrl_alerting_repo }}"
+    state=directory
+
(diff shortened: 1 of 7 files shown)


roles/tendrl-alerting/tasks/source.yml
```diff
@@ -0,0 +1,51 @@
...
+    chdir={{ tendrl_alerting_repo }}
+
+- name: Use example config files as initial configuration (see installation.rst)
+  copy:
+    src={{ item.src }}
+    dest=/etc/tendrl/{{ item.dest }}
+    remote_src=True
+  with_items:
+    - src: "{{ tendrl_alerting_repo }}/etc/tendrl/tendrl.conf.sample"
+      dest: 'tendrl.conf'
```
mbukatov: Are we sure we would like to do this? The `tendrl.conf` file is shared across components and so this step would remove any previous chagnes.

PR #23, opened 2017-01-11
Add role for installing ceph-ansible and its confiuration

mkudlej:


mkudlej:
@Tendrl/qe please review

roles/ceph-ansible/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+# list of devices which will be used for Ceph.
+# If it is empty, list is autodetected by ceph-ansible
+devices: []
+
+# use collocated journals? Default to true for simplicity.
+collocation: true
+
+# if collocation is false and journal devices are set,
+# create journals on journal devices.
```
dahorak: What if `collocation` is `false` but `journal_devices` are not set? Does it use some _autodetection_, same as for empty `devices`, or it is not supported configuration?

roles/ceph-ansible/tasks/main.yml
```diff
@@ -0,0 +1,138 @@
...
+    line: 'calamari: true'
+
+# edit osds.yml file
+
+- name: Uncomment devices
+  lineinfile:
+    dest: /usr/share/ceph-ansible/group_vars/osds.yml
+    regexp: '^#devices:'
+    line: 'devices:'
+  when: "{{ devices }}"
```
dahorak: The double curly brackets `{{` and `}}` shouldn't be required here ([see the Ansible doc](http://docs.ansible.com/ansible/playbooks_conditionals.html#the-when-statement)).

roles/ceph-ansible/tasks/main.yml
```diff
@@ -0,0 +1,138 @@
...
+    line: "  - /dev/{{item}}"
+  with_items: "{{ devices }}"
+  when: "{{ devices }}"
+
+- name: Use device autodetection if devices are not set
+  lineinfile:
+    dest: /usr/share/ceph-ansible/group_vars/osds.yml
+    regexp: "^#osd_auto_discovery: false"
+    line: "osd_auto_discovery: true"
+  when: "{{ not devices }}"
```
dahorak: Does this construction works correctly? (maybe yes, I'm just asking, because the double quotes and double curly brackets are not usually used in such `when` construction.

roles/ceph-ansible/tasks/main.yml
```diff
@@ -0,0 +1,138 @@
...
+  when: "{{ (not collocation) and journal_devices }}"
+
+- name: Uncomment journal devices
+  lineinfile:
+    dest: /usr/share/ceph-ansible/group_vars/osds.yml
+    regexp: '^#raw_journal_devices:'
+    line: 'raw_journal_devices:'
+  when: "{{ (not collocation) and journal_devices }}"
+
+- name: Add devices for if they are set
```
dahorak: I think, this should be:
_Add devices for **journal** if they are set_

roles/ceph-ansible/defaults/main.yml
mkudlej: If ``devices`` variable is empty devices are autodetected by ceph-ansible. If ``collocation`` is ``false``,  ``journal_devices`` should contain list of devices for journals.

roles/ceph-ansible/tasks/main.yml
mkudlej: Yes, it works.

roles/ceph-ansible/tasks/main.yml
mkudlej: It works with double curly brackets.

roles/ceph-ansible/tasks/main.yml
mkudlej: You're right, I'll fix it.

roles/ceph-ansible/tasks/main.yml
```diff
@@ -0,0 +1,138 @@
+---
+# ceph-ansible package is not in Epel yet. This playbook expects that
+# ceph-ansible is preinstalled or available in enabled yum repositories.
+# TODO do apt installation
+- name: Install ceph-ansible package
+  yum:
+    name=ceph-ansible
```
mbukatov: I see that you just install `ceph-ansible` package without enabling particular repository for it. That is good.

That said, I would like to know which ceph-ansible package you tested this role with?

mbukatov:
I would suggest to move installation of `ceph-ansible` package into qe-server role, and then enable repo with this package in qe_server playbook.

mbukatov:
Moreover it would be nice if it would be possible to run ceph-ansible playbooks via [pytest ansible playbook plugin](https://pypi.python.org/pypi/pytest-ansible-playbook).

mbukatov:
I'm dumping my notes here:

* we are not testing ceph-ansible, we are just using it
* we use it for setup of import cluster test case(s)
* ceph-ansible will be installed, configured and used on qe_server
* we install it from rpm package (which has a minor issue right now, because the package is not available in Storage SIG)
* we have ansible role for configuration of ceph ansible (see tasks in this pull request)

mkudlej:
@dahorak @mbukatov please review my changes

roles/ceph-ansible/README.rst
```diff
@@ -0,0 +1,14 @@
+===================================
+ Setup ceph-ansible to deploy Ceph
+===================================
+
+This role setups Ceph-ansible to be able to install Ceph.
+Role can be configured by variables mentioned in ``defaults/main.yml``.
+As it is written in ``tasks/main.yml`` ceph-ansible package should be 
+preinstalled or available in enabled yum repository.
+
+Example usage of configured ceph-ansible:
```
dahorak: Following lines sounds slightly strangely (or maybe I just misunderstand them). Could you please check it and at least fix the format, so it will be properly generated? (check the View button)

roles/ceph-ansible/defaults/main.yml
```diff
@@ -0,0 +1,19 @@
+---
+# home directory which contains copy of files in ceph-ansible package
+home_dir: "{{ ansible_env['WORKSPACE'] | default(ansible_env['HOME']) }}"
```
dahorak: The name ``home_dir`` sounds to me, like it should be the HOME dir of the user, I would name it maybe something like ``base_dir`` (if I understand the meaning correctly).
Also where should point the ``ansible_env['WORKSPACE']``? (I wasn't able to find, what it means.)

roles/ceph-ansible/defaults/main.yml
```diff
@@ -0,0 +1,19 @@
...
+collocation: true
+
+# if collocation is false and journal devices are set,
+# create journals on journal devices.
+# Default is empty because collocation is true.
+# collocation == false and len(journal_devices) == 0 is not supported!
+journal_devices: []
+
+# path to ansible configuration which overrides default ansible configuration
+ansible_conf: /tmp/ansible.cfg
```
dahorak: I'm not sure, if `/tmp/` is the best place for the configuration file (it might not persist reboot of the machine, but this didn't apply to RHEL 7/CentOS 7 though).

roles/ceph-ansible/tasks/main.yml
```diff
@@ -0,0 +1,180 @@
...
+    regexp: '^#cluster: ceph.*'
+    line: 'cluster: test_name'
+
+- name: Set automated fsid generation
+  lineinfile:
+    dest: "{{ home_dir }}/ceph-ansible/group_vars/all.yml"
+    regexp: '^#generate_fsid: true.*'
+    line: 'generate_fsid: true'
+
+# Disabled because this list in directly in ceph-ansible
```
dahorak: Probably just typo?  _Disabled because this list **is** directly in ceph-ansible._

roles/ceph-ansible/tasks/main.yml
```diff
@@ -0,0 +1,180 @@
...
+- name: Set public_network to ansible_default_ipv4['address']/16
+  lineinfile:
+    dest: "{{ home_dir }}/ceph-ansible/group_vars/all.yml"
+    regexp: "^#public_network: 0.0.0.0/0.*"
+    line: "public_network: {{ ansible_default_ipv4['address'] }}/16"
+
+# edit mons.yml file
+
+# Disabled on CentOS because calamari is not available in SIG
+- name: Enable calamari on mons
```
dahorak: Just question, it will be installed, configured and running on all mons, or only on one of them?

roles/qe-server/README.rst
```diff
@@ -3,3 +3,78 @@
 ================
 
 This role configures the machine for execution of usmqe integration tests.
+
```
dahorak: It seems like this and the following files are part of this pull request by mistake as it is about _Xvfb_ server and other stuff and not about _ceph-ansible_.

roles/ceph-ansible/defaults/main.yml
mkudlej: WORKSPACE is variable in Jenkins environment for build workspace, right?

roles/ceph-ansible/README.rst
mkudlej: Corrected.

roles/ceph-ansible/defaults/main.yml
dahorak: @mkudlej ah, yes you are right. Please, write there short description about that and consider the renaming of the ``home_dir`` to something like ``base_dir`` or ``work_dir``.

roles/ceph-ansible/defaults/main.yml
mkudlej: I think that is not problem, because it is written every time you run this playbook. This is not configuration file which should persist after reboot.

roles/ceph-ansible/tasks/main.yml
mkudlej: Done

roles/ceph-ansible/tasks/main.yml
mkudlej: Ceph-ansible install and run calamari on all mons. Console 2 has no problem with this. It is OK to create user in Calamari only on one instance after installation for import purpose.

mkudlej:
Tested with ceph-ansible https://koji.fedoraproject.org/koji/buildinfo?buildID=831491

roles/ceph-ansible/defaults/main.yml
dahorak: Ok, I agree (I somehow thing it is launched during the qe server configuration).

roles/ceph-ansible/README.rst
```diff
@@ -8,7 +8,11 @@ As it is written in ``tasks/main.yml`` ceph-ansible package should be
...
-# cd to ceph-ansible dir so relative paths can be used
-$ cd ``{{ home_dir }}/ceph-ansible from defaults/main.yml``
-# /tmp/ansible.cfg is from {{ ansible_conf }} from defaults/main.yml
-$ ANSIBLE_CONFIG=/tmp/ansible.cfg ansible-playbook -i ~/ceph_inventory site.yml
+
+.. code:: bash
+
+  # cd to ceph-ansible dir so relative paths can be used
+  # default value of variable $base_dir is in defaults/main.yml
+  $ cd $base_dir/ceph-ansible from defaults/main.yml
```
dahorak: Thanks, now it make sense. Maybe just remove the ``from defaults/main.yml`` from the command.

install_ceph.yml
```diff
@@ -0,0 +1,6 @@
+---
```
mbukatov: Could we rename this file to just `ceph_ansible.yml` or `configura_ceph_ansible.yml`?

PR #28, opened 2017-01-13
Setup for selenium web tests

mbukatov:
This pull requests adds the following tasks into usmqe role:

* installation of Xvfb and x11vnc systemd services
* opening of 5900 tcp port

Resolves https://github.com/Tendrl/usmqe-setup/issues/24


mbukatov:
@dahorak This is not ready to be merged, we may tweak the role a bit first. Eg. we may not like to open vnc port by default. But it already works, see the [updated readme file of qe-server role](https://github.com/mbukatov/usmqe-setup/blob/953c0763bc524f963d9e00e32594bb7457f4c6f7/roles/qe-server/README.rst).

roles/qe-server/README.rst
```diff
@@ -3,3 +3,72 @@
...
+.. _`Xvfb server`: https://en.wikipedia.org/wiki/Xvfb
+.. _`systemd template unit file`: https://fedoramagazine.org/systemd-template-unit-files/
+
+
+x11vnc Service
+==============
+
+Another systemd service installed in this role is ``x11vnc@.service``.
+
+Also note that the role opens tcp port ``5900`` for the vnc server.
```
dahorak: If we will use more than one DISPLAY and we would like to start VNC server for each of them, we will need more than the one port (``5900``).
My suggestion is, to use following ports (``5901, 5902, 5903,...``) as it is common for example for libvirt VMs.
The best way to determine the port number would probably be ``5900`` + ``DISPLAY number``. (it would be clearer than randomly assigned ports.)

roles/qe-server/files/x11vnc@.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=x11vnc allows VNC connections to real X11 displays
+# TODO: it's not possible to reference parametrized requirements like that
+# Requires=Xvfb@%I.service
+# After=Xvfb@%I.service
+
+[Service]
+Type=forking
+ExecStart=/usr/bin/x11vnc -display %I -bg -nopw -xkb -forever -shared -logfile /var/log/x11vnc.%I.log
```
dahorak: As I mentioned above, would it be possible to specify also the port number determined from the display number?

roles/qe-server/tasks/main.yml
```diff
@@ -49,6 +49,37 @@
...
+    owner=root
+    group=root
+    setype=systemd_unit_file_t
+  with_items:
+    - Xvfb@.service
+    - x11vnc@.service
+
+- name: Open port for x11vnc server via firewalld
+  firewalld:
+    port=5900/tcp
```
dahorak: Not sure, how to best solve the various port numbers here so I would maybe open range of ports ``5900 - 5920`` (or more/less, not sure how many opened displays are likely possible).

roles/qe-server/README.rst
mbukatov: This is valid point, the current state doesn't account for that.

mbukatov:
Based on discussion with @dahorak , I'm going to remove template feature from service ini file, specify dependencies properly, start both services and keep the firewall port 5900 open.

dahorak:
LGTM

PR #29, opened 2017-01-16
update links and names for tendrl-api role

fbalak:


roles/tendrl-api/README.rst
```diff
@@ -2,7 +2,7 @@
  Tendrl-api
 ============
 
-This role installs `tendrl-api`_ component.
+This role installs `api`_ component.
```
mbukatov: I would keep tendrl-api there.

roles/tendrl-api/README.rst
```diff
@@ -19,7 +19,7 @@ Tendrl`_ documentation with the following exceptions:
   configuration`_
 
 
-.. _`tendrl-api`: https://github.com/Tendrl/tendrl-api
-.. _`README file`: https://github.com/Tendrl/tendrl-api/blob/master/README.adoc
+.. _`api`: https://github.com/Tendrl/api
```
mbukatov: Name should remain, change only the url.

roles/tendrl-api/tasks/source.yml
```diff
@@ -81,14 +81,14 @@
     value="/usr/local/bin/puma -e production --dir {{ tendrl_api_repo }}"
 
 # as documented in README file and deployment.adoc from Tendrl/documentation
-- name: Apache config file for tendrl-api to run on port 80
```
mbukatov: I would keep the name

PR #30, opened 2017-01-17
comment copy sample config

fbalak:
installation is changed, now there is no conf sample in tendrl-commons


mbukatov:
@fbalak could you add link to documentation and/or link to commit which does this in the commons side?

mbukatov:
btw, I would rather drop the entire block instead of commenting it out - we can always restore deleted lines from git

fbalak:
link: https://github.com/Tendrl/commons/commit/bed2c7a840bbe73bf08dd3b2d287406fd441c067#diff-423eaed28c1d057218887cb6e5695734

PR #32, opened 2017-01-30
add basic playbook for testing correct instalation of tendrl

fbalak:


mkudlej:
@fbalak I agree with @mbukatov. So I'm closing this and I'll open issue in usmqe-tests.

PR #36, opened 2017-02-06
Add qe_pre_installation_tasks.yml playbook for auxiliary pre-installation tasks, such as package update.

dahorak:
Added qe_pre_installation_tasks.yml playbook for such auxiliary pre-installation tasks, such as package update.

### DIFF
--- a/qe_pre_installation_tasks.yml
+++ b/qe_pre_installation_tasks.yml
@@ -1,0 +1,12 @@
+---
+#
+# Perform auxiliary tasks before installing Tendrl itself:
+# * update all packages
+#
+- hosts: all
+  remote_user: root
+  tasks:
+    - name: upgrade all packages
+      yum:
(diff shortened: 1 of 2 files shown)


qe_pre_installation_tasks.yml
```diff
@@ -0,0 +1,12 @@
+---
+#
+# Perform auxiliary tasks before installing Tendrl itself:
+# * update all packages
+#
+- hosts: all
+  remote_user: root
+  tasks:
+    - name: upgrade all packages
+      yum:
```
mkudlej: Please consider also Ubuntu testing.

dahorak: No part of the setup playbooks count with Ubuntu now, so we will have to solve it globally. Thus I don't think it make much sense to add some _if_ to this one step now.

PR #40, opened 2017-02-14
Ceph installation

dahorak:
- Prepare environment for ceph.
- fix issues with ceph-ansible role

### DIFF
--- a/ceph_ansible.yml
+++ b/ceph_ansible.yml
@@ -1,4 +1,10 @@
 ---
+- hosts: ceph_mon
+  remote_user: root
+  tasks:
+    - name: gather facts from Ceph Mon nodes
+      setup:
+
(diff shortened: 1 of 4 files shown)


ceph_ansible.yml
```diff
@@ -1,4 +1,10 @@
 ---
+- hosts: ceph_mon
```
fbalak: Maybe this should be moved to /meta/main.yml under dependencies.

dahorak: @fbalak I'm thinking about using dependencies, but I'm not sure if you can use anything else than role as dependency and execute it on different host, than the main role.

fbalak: Ok, I didn't find a better way to do that.

PR #41, opened 2017-02-15
Split tendrl_gluster.yml to separated playbooks for tendrl and glusteror ceph

dahorak:
And add default playbook for CI environment.

### DIFF
--- a/gluster.yml
+++ b/gluster.yml
@@ -1,0 +1,10 @@
+---
+#
+# Install gluster
+#
+- hosts: gluster
+  remote_user: root
+  vars:
+    install_from: packages
+  roles:
+    - gluster-centos
--- a/tendrl_gluster_peers_bricks.yml
+++ b/tendrl_gluster_peers_bricks.yml
@@ -1,9 +1,10 @@
 ---
 #
-# * Install Tendrl server and tendrl-gluster nodes.
-# * Configure gluster trusted storage pool and create bricks 
-#   on all available devices on gluster nodes.
+# Install Tendrl server and tendrl-gluster nodes.
+# Configure gluster trusted storage pool and create bricks
+# on all available devices on gluster nodes.
 #
 - include: qe_pre_installation_tasks.yml
-- include: tendrl_gluster.yml
+- include: tendrl_server.yml
+- include: gluster.yml
 - include: gluster_peers_bricks.yml
--- a/tendrl_gluster_peers_bricks.yml
+++ b/tendrl_gluster_peers_bricks.yml
@@ -6,5 +6,7 @@
 #
 - include: qe_pre_installation_tasks.yml
 - include: tendrl_server.yml
+- include: tendrl_node.yml
+- include: tendrl_gluster.yml
 - include: gluster.yml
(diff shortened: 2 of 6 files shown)


tendrl_gluster_peers_bricks.yml
```diff
@@ -1,9 +1,10 @@
...
 #
-# * Install Tendrl server and tendrl-gluster nodes.
-# * Configure gluster trusted storage pool and create bricks 
-#   on all available devices on gluster nodes.
+# Install Tendrl server and tendrl-gluster nodes.
+# Configure gluster trusted storage pool and create bricks
+# on all available devices on gluster nodes.
 #
 - include: qe_pre_installation_tasks.yml
-- include: tendrl_gluster.yml
```
fbalak: Just a question. Why is `tendrl.yml` and `gluster.yml` used instead of `tendrl_gluster.yml` that also includes them? Is there some reason for keeping `tendrl_gluster.yml` playbook? Maybe if you would drop this playbook then the setup infrastructure would be simpler...

dahorak: My opinion is, that `tendrl_gluster.yml` playbook is redundant and should be removed, but I would like to keep it there for few days for backward compatibility with jenkins jobs.

fbalak: Ok.

mbukatov: I would rather have a single playbook, clearly maintained in git repo, for each jenkins job. So I'm in favor of keeping `tendrl_gluster.yml` playbook.

gluster.yml
```diff
@@ -0,0 +1,17 @@
+---
```
mbukatov: I have a problem with name of this playbook file: name `gluster.yml` suggests that this is concerned with gluster only, but it also installs tendrl components for gluster storage. For this reason, could we rename this to `tendrl_gluster_server.yml` (or something similar)?

mbukatov: I would like to clearly distinguish between playbooks which deals with gluster components only (eg. `gluster_peer_bricks.yml` and those which handles both gluster and tendrl).

tendrl.yml
```diff
@@ -0,0 +1,15 @@
+---
```
mbukatov: Similar, less serious note about naming here: would it make sense to rename it to `tendrl_server`?

dahorak: Agree, it make sense.

gluster.yml
dahorak: good point, now I'm thinking about splitting `gluster.yml` into two playbooks: one just for Gluster and second for Tendrl related things on storage nodes (node-agent, performance-monitoring, gluster-integration, ceph-integration...)

dahorak:
@mbukatov I've split the tasks to separated playbooks for Tendrl related stuff (`tendrl_server`, `tendrl_node`, `tendrl_gluster` and `tendrl_ceph`) and Gluster/Ceph related things (`gluster.yml`).
Default playbook for CI/automated installation is ci_default.yml - combining everything together.

PR #47, opened 2017-02-28
Fix ceph ansible local

mbukatov:
This pull request fixes https://github.com/Tendrl/usmqe-setup/issues/43 by allowing usmqe user to become root via sudo and using this power to execute `ceph-ansible` role locally under root user.

### DIFF
--- a/ceph_ansible.yml
+++ b/ceph_ansible.yml
@@ -6,7 +6,8 @@
       setup:
 
 - name: Install ceph-ansible and install Ceph
-  hosts: qe_server
+  hosts: localhost
+  connection: local
   user: root
   roles:
     - ceph-ansible
--- a/ceph_ansible.yml
+++ b/ceph_ansible.yml
@@ -8,6 +8,5 @@
 - name: Install ceph-ansible and install Ceph
   hosts: localhost
   connection: local
-  user: root
   roles:
     - ceph-ansible


roles/qe-server/tasks/main.yml
```diff
@@ -16,6 +16,14 @@
     comment="USM QE test user account"
     state=present
 
+- name: Allow usmqe user to do anything via sudo
+  copy:
+    src=20_usmqe_sudoers
+    dest=/etc/sudoers.d/20_usmqe_sudoers
```
mkudlej: Can you install this as usmqe user?

mbukatov: No you can't and it's ok: the qe-server role is expected to be executed from remote machine (eg. one's laptop or jenkins slave) during qe server setup, remotely as a root. Is that what you are asking about?

ceph_ansible.yml
```diff
@@ -6,7 +6,10 @@
       setup:
 
 - name: Install ceph-ansible and install Ceph
-  hosts: qe_server
-  user: root
+  hosts: localhost
+  connection: local
+  become: yes
+  become_user: root
+  become_method: sudo
```
dahorak: It is not necessary (and recommended) to run `ceph-ansible` role as `root`, it should be executed under the same user who launch the whole playbook. So the above three lines (11-13) are should not be there.

roles/qe-server/files/20_usmqe_sudoers
```diff
@@ -0,0 +1 @@
+usmqe ALL = (ALL) NOPASSWD: ALL
```
dahorak: Because of the above mentioned note, I'm not sure if this is necessary.

roles/qe-server/tasks/main.yml
```diff
@@ -16,6 +16,14 @@
...
+    mode=440
+    owner=root
+    group=root
```
dahorak: The same as above.

ceph_ansible.yml
dahorak: I've tested it and it should work this way:
```
- name: Install ceph-ansible and install Ceph
  hosts: localhost
  connection: local
  roles:
    - ceph-ansible
```

mbukatov:
@dahorak I updated the pull request.

PR #49, opened 2017-03-01
WORKAROUND qe_pre_installation_tasks: configure selinux to permissive mode

dahorak:
https://github.com/Tendrl/documentation/wiki/Tendrl-Package-Installation-Reference/60ab08c13ec3240d247c71c0c6c9690ef8a47ecd

### DIFF
--- a/qe_pre_installation_tasks.yml
+++ b/qe_pre_installation_tasks.yml
@@ -10,3 +10,7 @@
       yum:
         name: '*'
         state: latest
+    - name: WORKAROUND configure selinux to permissive mode
+      selinux:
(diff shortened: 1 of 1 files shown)


mbukatov:
Looks good. That said, could you link the reason for this change either into description of this pull request of in the commit message?

mbukatov:
Also, when this pull request is merged, we should create an issue for usmqe-setup titled "Remove Workaround: SELinux should not be disabled" which would link to this particular commit, so that we could track workarounds in usmqe-setup (it would be easily filterable in list of issues) and revertable.

PR #54, opened 2017-03-07
usmqe sudoers configuration

mbukatov:
This pull request configures sudo to allow usmqe group to do anything as root via sudo.


mbukatov:
Based on https://github.com/Tendrl/usmqe-setup/pull/47#pullrequestreview-25188851

PR #57, opened 2017-03-09
workaround for https://github.com/Tendrl/api/issues/86

fbalak:
based on:
https://github.com/Tendrl/api/issues/86#issuecomment-285063914
TODO: remove this after issue resolution

### DIFF
--- a/ci_default.yml
+++ b/ci_default.yml
@@ -25,3 +25,22 @@
   when: groups.ceph_osd is defined and groups.ceph_mon is defined
 - include: tendrl_ceph.yml
   when: groups.ceph_osd is defined and groups.ceph_mon is defined
+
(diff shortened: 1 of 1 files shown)


ci_default.yml
```diff
@@ -25,3 +25,23 @@
...
+
+    # immediate restart doesn't work
+    - pause:
+        seconds: 20
+  
+    - name: Start and enable tendrl-node-agent daemon
+      service:
+        name=tendrl-node-agent
+        state=started
+        enabled=yes
```
mbukatov: Would tendrl-node-agent be already enabled at this point? If this workaround is only about doing stop, sleep and start again sequence, adding `enabled=yes` may not be good idea.

PR #60, opened 2017-03-13
qe_reboot playbook

dahorak:
Adding playbook for rebooting all machines (usable mainly after machine upgrade).

### DIFF
--- a/ci_default.yml
+++ b/ci_default.yml
@@ -10,6 +10,8 @@
 # * Preconfigure ceph-ansible.
 - include: qe_pre_installation_tasks.yml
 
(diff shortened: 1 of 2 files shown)


mbukatov:
It may be related to the fact that `atd` is not running on my machines:

```
# systemctl status atd
● atd.service - Job spooling tools
   Loaded: loaded (/usr/lib/systemd/system/atd.service; enabled; vendor preset: enabled)
   Active: inactive (dead)
```

dahorak:
@mbukatov Good catch!

dahorak:
Should be fixed now.

PR #61, opened 2017-03-14
add option for tendrl-devel repo

fbalak:


roles/tendrl-repo/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
```
mbukatov: Could we refer to the development builds as develop and not devel? Something like `s/devel/develop/` in this patch would be required. I would rather use the exact same name as used for both copr builds and branch.

roles/tendrl-repo/tasks/main.yml
```diff
@@ -1,9 +1,23 @@
...
-- name: Add public key
+- name: Add public key for master repo
   rpm_key:
     key="https://copr-be.cloud.fedoraproject.org/results/tendrl/tendrl/pubkey.gpg"
+  when: enable_master_repo
+
+- name: Add tendrl devel repo
+  yum_repository:
+    name: tendrl-devel
+    description: Copr repo for tendrl owned by tendrl
```
mbukatov: It would help if we add `develop` int the description as well.

roles/tendrl-repo/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
...
+# add yum repository and pblic key for master tendrl repo
```
ltrilety: s/pblic/public

roles/tendrl-repo/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
...
+enable_master_repo: true
+# add yum repository and pblic key for develop tendrl repo
```
ltrilety: s/pblic/public

PR #62, opened 2017-03-15
Separated playbook for installation of usefull tools.

dahorak:
For example:
- bind-utils
- git
- psmisc
- tmux
- tree
- vim

### DIFF
--- a/qe_pre_installation_tasks.yml
+++ b/qe_pre_installation_tasks.yml
@@ -2,6 +2,7 @@
 #
 # Perform auxiliary tasks before installing Tendrl itself:
 # * update all packages
+# * WORKAROUND: set selinux to permissive mode
 #
 - hosts: usm_server:usm_nodes
   remote_user: root
--- a/qe_tools.yml
+++ b/qe_tools.yml
@@ -1,0 +1,19 @@
+---
+#
+# Install useful and auxiliary tools
+#
+- hosts: usm_server:usm_nodes
+  remote_user: root
+  tasks:
+    - name: Install useful tools
+      yum: name={{ item }} state=installed
+      with_items:
+        - bash-completion
+        - bind-utils
+        - git
+        - psmisc
+        - screen
+        - tmux
+        - tree
+        - vim
+      ignore_errors: yes


dahorak:
@mkudlej How do you mean that and when the playbook should/might be launched?
Could it be part of the automated installation, or should it be installed only on user request?

My idea here was, that it is quite separated playbook and neither of the installed package should influence the testing (I think, that those packages are usually part of typical system installation).

mkudlej:
@dahorak I think installation of these packages should be done by user and not by automated installation. I think it will be bad if any of these packages or their dependencies will influence our CI.

qe_pre_installation_tasks.yml
```diff
@@ -10,6 +10,17 @@
...
+      yum: name={{ item }} state=installed
+      with_items:
+        - bash-completion
+        - bind-utils
+        - git
+        - psmisc
+        - tmux
+        - tree
+        - vim
+      ignore_errors: yes
```
mbukatov: Would it make sense to have this in a separate playbooks (something like `qe_tools.yml`) so that I can rerun it separately when needed?

dahorak:
@fbalak In this case I would vote for separated playbok, than another variable/switch in jenkins...

mbukatov:
> I think installation of these packages should be part of separated playbook to avoid influence of testing.

All those packages are from core repositories, so I don't think that it could break anything or tainted test environment.

dahorak:
Well, I see one possible problem - if some Tendrl component will depends on one of the tool (or some dependent package) and the dependency will not be properly set in the package, we might miss it, because it will be fulfilled... But I'm not sure how likely it is...

dahorak:
I moved the installation to separated playbook `qe_tools.yml`. Now the question is, if it is ok to install it automatically or not?

mbukatov:
@dahorak That's good point, maybe we should not include this in `qe_pre_installation_tasks.yml`, keeping `qe_tools.yml` in separate playbook which one could run easily later when needed.

ltrilety:
What about to create a related task for this playbook in jenkins, that will be appreciated I think.

dahorak:
@ltrilety I'm afraid, that new job in jenkins for each similar task will make it slightly obfuscated :-/

dahorak:
I've got an Idea, how to do it with the jenkins job - I don't want to create new job for each cluster, but it might be possible to create just one job and configure which task you want to perform and on which cluster...

dahorak:
I've got an Idea, how to do it with the jenkins job - I don't want to create new job for each cluster, but it might be possible to create just one job and configure which task you want to perform and on which cluster... (just idea for future feature...)

dahorak:
@mkudlej what do you think about this now? (when it is in new separated playbook)

PR #63, opened 2017-03-20
Install rpmlint and rpmdeplint for package validation tasks.

dahorak:


dahorak:
@mbukatov are you ok with installation of those two packages (`rpmlint` and `rpmdeplint`) directly in `qe_server` role?

PR #64, opened 2017-03-22
create module and basic playbook for creating admin user

fbalak:


library/tendrl_user.py
```diff
@@ -0,0 +1,100 @@
...
+    with default password."""
+
+    password = re.compile("\"Password:\s*(?P<pass>[a-zA-Z0-9\._]+)\"")
+    admin_exists = re.compile("User named admin already exists.")
+    password_found = password.search(string)
+    admin_found = admin_exists.search(string)
+    if password_found is not None:
+        return password_found.group("pass")
+    elif admin_found is not None:
+        login("admin", new_password)
```
mbukatov: Ah, this `login()` may end the whole process on failure. Could you drop a comment bout that?

library/tendrl_user.py
```diff
@@ -0,0 +1,100 @@
...
+
+def generate_user():
+    """Generates admin according to:
+    https://github.com/Tendrl/api/blob/master/docs/users.adoc#create-admin-user"""
+
+    os.chdir("/usr/share/tendrl-api")
+    return os.popen("RACK_ENV=production rake etcd:load_admin").read()
+
+
+def parse_password(string, module):
```
mbukatov: I'm little puzzled why sometimes you are massing this module variable and sometimes you are just using it as a global variable.

library/tendrl_user.py
fbalak: This login tests if admin that already exists can log in with password admin123. If he can log in then task stays green. If he can not login then task is red and in output is seen that he tried to log in with username admin and password admin123.

library/tendrl_user.py
fbalak: My bad, I will remove the parameter.

library/tendrl_user.py
```diff
@@ -0,0 +1,99 @@
...
+import os
+import socket
+import json
+
+hostname = socket.gethostname()
+module = AnsibleModule(argument_spec={})
+new_password = "admin123"
+
+
+def generate_user():
```
mkudlej: I think loading default admin should be in usmqe-setup repo as ansible code

library/tendrl_user.py
```diff
@@ -0,0 +1,99 @@
...
+        response = {"url": request.url, "data": post_data, "token": token}
+        module.fail_json(msg="Request to server failed.", meta=response)
+
+
+def main():
+    username = "admin"
+    user_info = str(generate_user())
+    password = parse_password(user_info)
+    token = login(username, password)["access_token"]
+    result = change_password(username, new_password, token)
```
mkudlej: I think that changing password for admin should be in separated ansible playbook just for manual testing.

library/tendrl_user.py
```diff
@@ -0,0 +1,99 @@
...
+    if request.status_code == requests.codes.ok:
+        return request.json()
+    else:
+        response = {"url": request.url, "data": post_data}
+        module.fail_json(
+            msg="Could not login with these credentials.",
+            meta=response)
+
+
+def change_password(username, password, token):
```
mkudlej: This is part of user tests and should not be here.

tendrl_admin.yml
```diff
@@ -0,0 +1,9 @@
+---
+- hosts: usm_server 
+  remote_user: root
+  tasks:
+    - name: Create user admin with password admin
```
ltrilety: hmm, and I thought that the password was set to admin123

fbalak: yes, this is typo

library/tendrl_user.py
mbukatov: I don't get this comment - *this is* ansible code.

mbukatov:
Ok, so after discussion with @mkudlej , it turned out that he is not opposing the idea of ansible module, but we would like to start with cleaner approach from the beginning: the functionality of this module would be split into 2:

* one for initialization of tednrl andmin password only, so that the autogenerated password is available in the playbook if needed
* one for changing password of given tendrl user

There are 2 main use cases:

* setup for automatic testing: playbook just runs 1st module and stores the password in the usm.ini file (via ini module)
* setup for manual tests: playbook runs 1st module and immediately changes the password to some default (hardcoded in playbook) via 2nd module

mkudlej:
This will be reworked.

mbukatov:
Note: work started here goes on in https://github.com/Tendrl/usmqe-setup/pull/67

PR #67, opened 2017-03-28
Admin user

mkudlej:
Tendrl-api role now loads default admin user.
tendrl_server playbook now creates usm.ini configuration file.
There is ansible module for changing password for user from usm.ini in library.
There is playbook tendrl_admin for changing password for admin user from usm.ini for manual testing.

### DIFF
--- a/library/tendrl_user.py
+++ b/library/tendrl_user.py
@@ -1,7 +1,8 @@
 #!/usr/bin/python
 '''
 module: test_user
-short_description: Changes 'usm_username' user password to 'admin123' via Tendrl API.
+short_description: Changes 'usm_username' user password to _password_from_module_argument_
+via Tendrl API.
 description:
     - Module reads from confile(module argument 'conf_path') 'usm_username',
       'usm_password' and 'usm_api_url'.
@@ -22,19 +23,21 @@ ANSIBLE_METADATA = {'metadata_version': '1.0',
 
 DOCUMENTATION = '''
 module: test_user
-short_description: Changes 'usm_username' user password to 'admin123' via Tendrl API.
+short_description: Changes 'usm_username' user password to _password_from_module_argument_
+via Tendrl API.
 description:
     - Module reads from confile(module argument 'conf_path') 'usm_username', 
       'usm_password' and 'usm_api_url'.
     - There must be installed tendrl-api on the machine where this is ran.
 version_added: "1.0"
+    - Usage: ansible -i _inventory_ -m tendrl_user -a "conf_path='_usm.ini.path_'" localhost
 author: "fbalak@redhat.com", "mkudlej@redhat.com"
 '''
 
-NEW_PASSWORD = "admin123"
 MODULE = AnsibleModule(
     argument_spec=dict(
-        conf_path=dict(required=True, type='str')
+        conf_path=dict(required=True, type='str'),
+        new_password=dict(required=True, type='str')
         )
     )
 
@@ -76,6 +79,7 @@ def main():
     """Main function"""
 
     conf_path = MODULE.params['conf_path']
+    new_password = MODULE.params['new_password']
     config = ConfigParser.SafeConfigParser()
     config.read(conf_path)
 
@@ -84,15 +88,17 @@ def main():
     api_url = config.get('usmqepytest', 'usm_api_url')
 
     token = login(api_url, username, password)["access_token"]
-    result = change_password(api_url, username, NEW_PASSWORD, token)
+    if password == new_password:
+        MODULE.exit_json(changed=False)
+
+    result = change_password(api_url, username, new_password, token)
     response = {"result": result}
 
-    config.set('usmqepytest', 'usm_password', NEW_PASSWORD)
+    config.set('usmqepytest', 'usm_password', new_password)
     with open(conf_path, 'wb') as conffile:
         config.write(conffile)
 
     MODULE.exit_json(changed=True, meta=response)
-
 
 if __name__ == '__main__':
     main()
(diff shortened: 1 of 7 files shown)


library/tendrl_user.py
```diff
@@ -0,0 +1,104 @@
...
+version_added: "1.0"
+author: "fbalak@redhat.com", "mkudlej@redhat.com"
+'''
+
+import json
+import ConfigParser
+import requests
+from ansible.module_utils.basic import AnsibleModule
+
+ANSIBLE_METADATA = {'metadata_version': '1.0',
```
fbalak: Python imports are placed after ansible special variables according to http://docs.ansible.com/ansible/dev_guide/developing_modules_documenting.html.

mkudlej: Solved

tendrl_admin.yml
```diff
@@ -0,0 +1,5 @@
+---
```
mbukatov: Could you change name of this playbook to `qe_tendrl_admin.yml`? So that it would be obvious that this is a qe playbook, not necessary for ci setup or not following official setup documentation.

tendrl_admin.yml
```diff
@@ -0,0 +1,5 @@
...
+- hosts: localhost
+  tasks:
+    - name: Change admin password to admin123
```
mbukatov: Could you note there that this module changes password of user specified in `conf_path` ini file as `usm_username`? Like this:

> Change password of usm admin user (specified as `usm_username` in  `conf_path` ini file) to admin123

library/tendrl_user.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/python
+'''
+module: test_user
```
mbukatov: This line should state: tendrl_user

library/tendrl_user.py
```diff
@@ -0,0 +1,104 @@
...
+short_description: Changes 'usm_username' user password to _password_from_module_argument_
+via Tendrl API.
+description:
+    - Module reads from confile(module argument 'conf_path') 'usm_username', 
+      'usm_password' and 'usm_api_url'.
+    - There must be installed tendrl-api on the machine where this is ran.
+version_added: "1.0"
+    - Usage: ansible -i _inventory_ -m tendrl_user -a "conf_path='_usm.ini.path_'" localhost
+author: "fbalak@redhat.com", "mkudlej@redhat.com"
```
mbukatov: Multiple authors are better specified like this:

```
author:
    - "fbalak@redhat.com"
    - "mkudlej@redhat.com"
```

(based on ansible core modules)

library/tendrl_user.py
```diff
@@ -0,0 +1,104 @@
...
+DOCUMENTATION = '''
+module: test_user
+short_description: Changes 'usm_username' user password to _password_from_module_argument_
+via Tendrl API.
+description:
+    - Module reads from confile(module argument 'conf_path') 'usm_username', 
+      'usm_password' and 'usm_api_url'.
+    - There must be installed tendrl-api on the machine where this is ran.
+version_added: "1.0"
+    - Usage: ansible -i _inventory_ -m tendrl_user -a "conf_path='_usm.ini.path_'" localhost
```
mbukatov: Usage should be [EXAMPLES block](https://docs.ansible.com/ansible/dev_guide/developing_modules_documenting.html#examples-block)

roles/tendrl-api/tasks/main.yml
```diff
@@ -37,3 +37,18 @@
...
+  args:
+    chdir: /usr/share/tendrl-api
+  register: rake_result
+  changed_when: 
+    - "'User named admin already exists.' not in rake_result.stdout"
+  failed_when:
+    - "rake_result.rc != 0"
+
+- name: Store password to file
+  shell: echo "{{rake_result.stdout_lines[3] | regex_replace('^.*\"Password:\s*(.*)\".*$','\1')}}" | tee /root/password
```
mbukatov: This will fail when you rerun the playbook again, because `Load admin user to etcd` task would be reported as "ok" and and so `rake_result` variable would be undefined. For this reason, it would be good idea to skip this as well when the previous task is not "changed".

tendrl_admin.yml
```diff
@@ -0,0 +1,5 @@
...
+      tendrl_user: conf_path="usm.ini" new_password="admin123"
```
mbukatov: Moreover I would suggest to remove `/root/password` file in this playbook to indicate that the default random password is no longer used. I don't see any value in modifying this file as well. What does @dahorak think?

library/tendrl_user.py
```diff
@@ -0,0 +1,104 @@
...
+        new_password=dict(required=True, type='str')
+        )
+    )
+
+def login(api_url, username, password):
+    """Login to tendrl server."""
+
+    post_data = {"username": username, "password": password}
+    request = requests.post(
+        "{}login".format(api_url),
```
mkudlej: Yes, but there should be slash. Otherwise will not work tests for API.

library/tendrl_user.py
```diff
@@ -0,0 +1,104 @@
...
+def change_password(api_url, username, password, token):
+    """Change password of user and set it's email to username@hostname"""
+
+    post_data = {
+        "username": username,
+        "password": password,
+        "password_confirmation": password,
+        "email": "{}@{}".format(username, 'localhost.localdomain')}
+    request = requests.put(
+        "{}users/{}".format(api_url, username),
```
mkudlej: Yes, but there should be slash. Otherwise will not work tests for API.

mbukatov: Ok.

library/tendrl_user.py
mbukatov: Ok.

PR #69, opened 2017-03-29
Firewall configuration for ceph cluster

dahorak:


dahorak:
Whole firewall configuration will be performed in slightly different way, so closing this pull request and creating new one.

PR #73, opened 2017-03-30
Firewall configuration

dahorak:
Firewall configuration for whole cluster moved to new playbook `tendrl_firewall.yml` (separated by particular role).
Workaround (disable firewall) moved to `tendrl_firewall_workaround.yml`.


mbukatov:
That said, later when we have a clear understanding of ceph/gluster vs tendrl firewall setup and related responsibilities and supported use cases, we may need to tweak or even change how we do the firewall configuration during setup (`ci_default.yml` playbook). That said, playbook `firewall_tendrl.yml` or firewall roles (such as `firewall-ceph-osd`) would remain as it is now.

PR #77, opened 2017-04-03
Separate Ceph and Gluster repos to new playbooks

dahorak:
We need to have separated playbooks or roles for configuring Ceph and Gluster repo, because Tendrl should be able to install Ceph and Gluster cluster itself but it requires the proper repositories configured.

### DIFF
--- a/roles/gluster-centos/defaults/main.yml
+++ b/roles/gluster-centos/defaults/main.yml
@@ -1,3 +1,0 @@
----
-# when true, tendrl-api is installed from a source repository
-gluster_repo_url: http://buildlogs.centos.org/centos/7/storage/x86_64/gluster-3.9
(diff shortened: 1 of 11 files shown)


roles/gluster-centos/defaults/main.yml
```diff
@@ -1,3 +1 @@
 ---
-# when true, tendrl-api is installed from a source repository
```
fbalak: Why is this file staying in `defaults` directory and is not removed along with the directory itself?

dahorak: My point here is, why remove file which might be needed later, so it's like place holder ... But I don't have strong opinion about this, so I have no problem with removing it.

fbalak: Ok, thanks. I personally would remove it because now it has no purpose and it can be recreated in the future but I don't mind having it there for now.

dahorak:
Ceph and Gluster repo related asks moved to separated roles.

PR #78, opened 2017-04-03
ci_tendrl_cluster.yml: playbook for installation of tendrl cluster without Ceph or Gluster cluster

dahorak:
This playbook perform following tasks, when applicable (e.g. there are nodes in gluster or ceph_* group)
* Configure firewall
* Install Tendrl server.
* Install Tendrl storage nodes.
* Add Gluster repo.
* Add Ceph repos.

### DIFF
--- a/ci_tendrl_cluster.yml
+++ b/ci_tendrl_cluster.yml
@@ -1,0 +1,58 @@
+---
+# Perform following tasks, when applicable
+# (e.g. there are nodes in gluster or ceph_* group)
+# * Configure firewall
+# * Install Tendrl server.
+# * Install Tendrl storage nodes.
+# * Add Gluster repo.
+# * Add Ceph repos.
+- include: qe_pre_installation_tasks.yml
+
+# configure firewall
+- include: firewall_tendrl.yml
+- include: firewall_tendrl_workaround.yml
+
+- include: qe_reboot.yml
+
+- include: tendrl_server.yml
+
+- include: tendrl_node.yml
+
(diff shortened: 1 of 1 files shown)


ci_tendrl_cluster.yml
```diff
@@ -0,0 +1,58 @@
...
+
+- include: tendrl_server.yml
+
+- include: tendrl_node.yml
+
+# Add Centos Gluster YUM repo
+- hosts: gluster
+  remote_user: root
+  roles:
+    - { role: gluster-centos-repo, when: groups.gluster is defined }
```
mbukatov: Is this condition `when: groups.gluster is defined` needed considering that you apply this role on members of this group in the first place?

ci_tendrl_cluster.yml
```diff
@@ -0,0 +1,58 @@
...
+
+# Install centos-release-ceph package with Ceph Repos
+- hosts: ceph_mon:ceph_osd
+  remote_user: root
+  roles:
+  - { role: ceph-centos-repo, when: groups.ceph_osd is defined and groups.ceph_mon is defined }
+
+# workaround according to https://github.com/Tendrl/api/issues/86#issuecomment-285063914
```
fbalak: I am going to test this issue. After that I will give comment if it can be removed.

ci_tendrl_cluster.yml
dahorak: No :) ... I'll remove it.

PR #82, opened 2017-04-06
tendrl-api: add new line to password file

dahorak:
The admin password is stored in `/root/password` file without the ending new line:
```
# cat password 
8b725666[root@usm-server ~]# 
```
This patch will add new line at the end of the password file.

### DIFF
--- a/qe_tendrl_admin.yml
+++ b/qe_tendrl_admin.yml
@@ -13,5 +13,5 @@
 
     - name: Change password in /root/password file
       copy:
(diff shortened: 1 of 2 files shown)


dahorak:
This is the result with this patch:
```
[root@usm-server ~]# cat password 
ed2a903d
[root@usm-server ~]#
```

dahorak:
@mbukatov good point, should be fixed now

PR #83, opened 2017-04-06
re-enable epel repository

mbukatov:
Since Tendrl itself finally starts to cover package installation to some
degree, we need to have epel enabled from now on for this to work.

Note: when you run these playbooks on RHEL, the setup fails - there is
no change in the check. This is a feature: enabling EPEL in such case
is a blocker bug and we need to catch it as soon as possible.

Based on @mkudlej feedback.


mbukatov:
I haven't tested this on a cluster (as I prefer not to rebuild it right now) ...

PR #85, opened 2017-04-13
add tendrl dependencies repo

fbalak:


mkudlej:
Absence of this breaks Tendrl installation, so I'm going to merge it and try to rebuild cluster. I expect that package sanity tests should be changed.

PR #86, opened 2017-04-19
add gdeploy repo

mkudlej:
This change is based on documentation change: https://github.com/Tendrl/documentation/wiki/Tendrl-Package-Installation-Reference/_compare/7251f8d8ba5e203953681f90438359a7ac90292c...e395971bb84c5b9308844cf9a5e0ee5c97516c43

### DIFF
--- a/ci_tendrl_cluster.yml
+++ b/ci_tendrl_cluster.yml
@@ -23,6 +23,7 @@
   remote_user: root
   roles:
     - gluster-centos-repo
+    - gluster-gdeploy-repo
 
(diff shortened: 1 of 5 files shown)


dahorak:
@fbalak _gdeploy_ is not part of Tendrl project, so it should not be part of the _tendrl-repo_ role.
But you are right that we should consider to add it to existing role... and I think [gluster-centos-repo](https://github.com/Tendrl/usmqe-setup/tree/master/roles/gluster-centos-repo) is good candidate.

mkudlej:
@dahorak please review

PR #91, opened 2017-04-26
retry failed package installation

dahorak:
Try each package installation via yum 5 times (if failing).

### DIFF
--- a/firewall_tendrl.yml
+++ b/firewall_tendrl.yml
@@ -9,6 +9,10 @@
       yum:
         name: 'firewalld'
         state: present
+      register: task_result
+      until: task_result|success
+      retries: 5
+      delay: 5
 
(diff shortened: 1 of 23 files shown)


dahorak:
> Does this qualify to be marked as a workaround? :)

From my point of view, it is some kind of "workaround", but I wouldn't mark it as workaround :)... I would consider it as "typical problem solving procedure" (_if it not work, try it again_). When I'm thinking about it, it might be good to add "yum clean all" between each try... but I'm not sure if it is possible to do it easily and straightforwardly in Ansible playbook.

> Doesn't one failure affect other yum tasks?

It shouldn't break anything, because the behavior from outside point of view should be the same as it is now (if it will fail 5 times in row, it will be considered as _failed_).

PR #92, opened 2017-05-04
Ceph-ansible add support for cluster_network

dahorak:
* configure second configured network as Ceph cluster_network (if available).
* calculate network prefix via jinja2 filter (instead of previously used `ipcalc` command)

### DIFF
--- a/roles/ceph-ansible/tasks/main.yml
+++ b/roles/ceph-ansible/tasks/main.yml
@@ -110,6 +110,37 @@
     regexp: "^#public_network: 0.0.0.0/0.*"
     line: "public_network: {{ pub_network }}/{{ pub_prefix }}"
 
+- name: Filter additional NICs (exclude ansible default interface and lo)
+  set_fact:
+    additional_network_interfaces: "{{ hostvars[groups['osds'][0]].ansible_interfaces | difference(['lo', hostvars[groups['osds'][0]].ansible_default_ipv4.interface]) }}"
+- name: Prepare empty list for additional_networks
+  set_fact:
+    additional_networks: []
+- name: Filter configured additional interfaces
+  set_fact:
+    additional_networks: "{{ additional_networks }} + [ {{ hostvars[groups['osds'][0]]['ansible_'+item].ipv4 }} ]"
+  when:
+   - "hostvars[groups['osds'][0]]['ansible_'+item] is defined"
+   - "hostvars[groups['osds'][0]]['ansible_'+item].ipv4 is defined"
+  with_items: "{{ additional_network_interfaces }}"
+- debug: var=additional_networks
+- name: Use first additional interface as data_network/data_prefix
+  set_fact:
+    data_network: "{{ additional_networks.0.network }}"
+    data_prefix: "{{ ['0.0.0.0/', additional_networks.0.netmask] | join('') | ipaddr('prefix') }}"
+  when: additional_networks.0 is defined
+- debug: var=data_network
+- debug: var=data_prefix
+
+- name: Set cluster_network to second NIC (if available on OSD nodes)
+  lineinfile:
+    dest: "{{ base_dir }}/ceph-ansible/group_vars/all.yml"
+    regexp: "^#cluster_network: .*"
+    line: "cluster_network: {{ data_network }}/{{ data_prefix }}"
+  when:
+    - "data_network is defined"
+    - "data_prefix is defined"
+
 # edit mons.yml file
 
 # Disabled on CentOS because calamari is not available in SIG
--- a/roles/ceph-ansible/tasks/main.yml
+++ b/roles/ceph-ansible/tasks/main.yml
@@ -100,16 +100,15 @@
     regexp: "^#journal_size: 0.*"
     line: 'journal_size: 5120'
 
-- name: calculate network prefix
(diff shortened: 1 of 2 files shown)


roles/ceph-ansible/tasks/main.yml
```diff
@@ -100,16 +100,46 @@
...
+- name: Set public_network to ansible_default_ipv4['network']/prefix
   lineinfile:
     dest: "{{ base_dir }}/ceph-ansible/group_vars/all.yml"
     regexp: "^#public_network: 0.0.0.0/0.*"
-    line: "public_network: {{ hostvars[groups['mons'][0]].ansible_default_ipv4['network'] }}/{{ net_prefix.stdout }}"
+    line: "public_network: {{ pub_network }}/{{ pub_prefix }}"
+
+- name: Filter additional NICs (exclude ansible default interface and lo)
+  set_fact:
+    additional_network_interfaces: "{{ hostvars[groups['osds'][0]].ansible_interfaces | difference(['lo', hostvars[groups['osds'][0]].ansible_default_ipv4.interface]) }}"
```
mbukatov: I see that you use `groups['osds'][0]` instead of previously used `groups['mons'][0]` to get the default interface. Is this intentional or would it be better to use the same expression to get it in both cases?

roles/ceph-ansible/tasks/main.yml
```diff
@@ -100,16 +100,46 @@
...
+  when:
+   - "hostvars[groups['osds'][0]]['ansible_'+item] is defined"
+   - "hostvars[groups['osds'][0]]['ansible_'+item].ipv4 is defined"
+  with_items: "{{ additional_network_interfaces }}"
+- debug: var=additional_networks
+- name: Use first additional interface as data_network/data_prefix
+  set_fact:
+    data_network: "{{ additional_networks.0.network }}"
+    data_prefix: "{{ ['0.0.0.0/', additional_networks.0.netmask] | join('') | ipaddr('prefix') }}"
+  when: additional_networks.0 is defined
```
mbukatov: > Use first additional interface as data network

Is this something incorporated into our Foreman deployment automation or would we need to tweak this in the future for some reason (eg. having particular interface for data)? It would be good to just write this down as a comment here, I'm not asking for extension here.

roles/ceph-ansible/tasks/main.yml
dahorak: The `cluster_network` is/might be available only on _osd_ nodes, so there should be `osds`. Maybe we could change the previously used `mons` to `osds`  too, but I think it is not necessary.

roles/ceph-ansible/tasks/main.yml
```diff
@@ -100,16 +100,46 @@
...
   lineinfile:
     dest: "{{ base_dir }}/ceph-ansible/group_vars/all.yml"
     regexp: "^#public_network: 0.0.0.0/0.*"
-    line: "public_network: {{ hostvars[groups['mons'][0]].ansible_default_ipv4['network'] }}/{{ net_prefix.stdout }}"
+    line: "public_network: {{ pub_network }}/{{ pub_prefix }}"
+
+- name: Filter additional NICs (exclude ansible default interface and lo)
+  set_fact:
+    additional_network_interfaces: "{{ hostvars[groups['osds'][0]].ansible_interfaces | difference(['lo', hostvars[groups['osds'][0]].ansible_default_ipv4.interface]) }}"
+- name: Prepare empty list for additional_networks
```
dahorak: yep, good point

roles/ceph-ansible/tasks/main.yml
mbukatov: But it's still worth it, as I wouldn't have to figure out if there actually is any reason for using `mons` vs `osds` on particular places. As I said, a readability improvement to make your intention more clear :)

roles/ceph-ansible/tasks/main.yml
dahorak: The code in this playbook is quite general - I've tested the particular code on few various scenarios with 1, 2, 3 and 5 NICs (or something like that), with various combinations of configured and not configured NICs and it seems to work quite smoothly and predictably. Only one exception is, when more than two NICs are configured, it is not possible to automatically distinguish which network should be the `data_network`, so the first one from the remainings is used.

About our current infrastructure, we are ready for testing with one and two networks (See the `SECOND_NETWORK` parameter when building new cluster). I've checked both installations and it works well.

roles/ceph-ansible/tasks/main.yml
```diff
@@ -100,16 +100,46 @@
...
-- name: calculate network prefix
-  shell: "ipcalc -p 0.0.0.0 {{ hostvars[groups['mons'][0]].ansible_default_ipv4['netmask'] }} | sed 's|PREFIX=||'"
-  register: net_prefix
-  changed_when: false
+- set_fact:
+    pub_network: "{{ hostvars[groups['mons'][0]].ansible_default_ipv4.network }}"
+    pub_prefix: "{{ ['0.0.0.0/', hostvars[groups['mons'][0]].ansible_default_ipv4.netmask] | join('') | ipaddr('prefix') }}"
 
-- name: Set public_network to ansible_default_ipv4['address']/16
+- name: Set public_network to ansible_default_ipv4['network']/prefix
```
dahorak: good point, fixed

PR #95, opened 2017-05-11
enable glusterd service

fbalak:
This adds enabling `glusterd` service after glusterfs is installed and `glusterd` started.


fbalak:
So we don't have to start glusterd service manually after restarting nodes. It is not done yet. https://gluster.readthedocs.io/en/latest/Administrator%20Guide/Start%20Stop%20Daemon/

PR #97, opened 2017-05-15
Remove installation from sources

fbalak:
Removes folders defaults and meta from roles that contain stuff related to installation from sources. Removes tasks/binary.yml and tasks/package.yml and contain of tasks/package.yml add to tasks/main.yml. Updates documentation for all roles and updates list of related issues for each role.

Resolves: https://github.com/Tendrl/usmqe-setup/issues/96


roles/tendrl-alerting/tasks/main.yml
```diff
@@ -2,11 +2,14 @@
...
+# TODO Add when packages are ready
+#- name: Install tendrl-alerting
+#  yum:
+#    name=tendrl-alerting
+#    state=latest
 
-- include: binary.yml
-  when: install_from == "packages"
+- fail: msg="There are no packages yet, we can't start and enable it"
+  when: not ignore_incomplete_installation
```
mbukatov: I guess that this is no longer true and `tendrl-alerting` package is available for some time. Which brings me to another detail: do we need this role when we don't seem to use it anymore?

roles/tendrl-ceph-integration/README.rst
```diff
@@ -2,27 +2,12 @@
...
-  via ansible metadata dependencies when ``install_from == source``)
-- we don't install `/etc/tendrl/tendrl.conf` configuration file, which is owned
-  by `tendrl-node-agent` component
-
-Open issues:
-
-* https://github.com/Tendrl/node_agent/issues/99
-* https://github.com/Tendrl/ceph_integration/issues/57
+This role should not be used in production playbooks because installation
+of this component is handled by `node-agent`_ as mentioned `here`_.
```
mbukatov: So do we still need this role?

roles/tendrl-commons/README.rst
```diff
@@ -2,12 +2,7 @@
  Tendrl-commons
 ===============
 
-This role installs `tendrl-commons`_ component.
-
-When variable ``install_from`` is set to ``packages``, the component is
-installed from rpm packages via ``yum``. On the other hand when the value is
-set to ``source``, tendrl-commons is installed from sources from github via
-``pip``.
+This role installs `tendrl-commons`_ component from packages.
```
mbukatov: Do we still need this role, when `tendrl-commons` will be installed as a dependency?

roles/tendrl-gluster-integration/README.rst
```diff
@@ -2,27 +2,12 @@
...
 
-This role installs `tendrl-gluster-integration`_ component.
+This role installs `tendrl-gluster-integration`_ component from packages.
 
-When variable ``install_from`` is set to ``packages``, the component is
-installed from rpm packages via ``yum``. On the other hand when the value is
-set to ``source``, tendrl-gluster-integration is installed from sources from
-github via ``pip``.
+This role should not be used in production playbooks because installation
+of this component is handled by `node-agent`_ as mentioned `here`_.
```
mbukatov: So do we still need `tendrl-gluster-integration` role?

roles/tendrl-commons/README.rst
```diff
@@ -2,12 +2,7 @@
...
 
 The installation is done according to `installation.rst`_ file with the
 following exceptions:
```
mbukatov: This and other lines which follows could be removed as well (describes source installation).

mbukatov:
Note to self/others: when merged, we need to update usmqe docs.

roles/tendrl-gluster-integration/README.rst
fbalak: Not really. It is kept there for emergency testing when provisioning is broken but we can remove it along with tendrl-ceph-integration.

roles/tendrl-commons/README.rst
fbalak: Probably not.

roles/tendrl-alerting/tasks/main.yml
fbalak: I would leave this role here. I will just update it and when they describe it better, we can use it and test it. According to [this](https://github.com/Tendrl/alerting/blob/develop/doc/source/installation.rst) it looks like it sends mails with reports and errors. But there is no mention in official [installation guide](https://github.com/Tendrl/documentation/wiki/Tendrl-Package-Installation-Reference).

PR #99, opened 2017-05-17
switch node-agent install task with cluster creation

fbalak:
Thanks to this change we should be able to import gluster cluster. For ceph cluster @dahorak needs to create workaround.

### DIFF
--- a/ci_default.yml
+++ b/ci_default.yml
@@ -19,8 +19,6 @@
 
 - include: tendrl_server.yml
 
-- include: tendrl_node.yml
-
 - include: gluster.yml
   when: groups.gluster is defined
 - include: gluster_peers_bricks.yml
@@ -28,6 +26,8 @@
 
 - include: ceph_prereq.yml
   when: groups.ceph_osd is defined and groups.ceph_mon is defined
+
+- include: tendrl_node.yml
 
 # workaround according to https://github.com/Tendrl/api/issues/86#issuecomment-285063914
 # TODO: delete after resolving


dahorak:
This issue doesn't apply to Ceph storage nodes, so there is no required workaround for that.

> (14:13:08) dahorak: rkanade: Hi, Filip mentioned, that we should firstly create Gluster or Ceph cluster on storage nodes and then install tendrl-node-agents, otherwise if we will firstly install and configure node-agents and then create the Gluster or Ceph storage, it will not work (which is probably the root cause of https://github.com/Tendrl/node-agent/issues/435). Did I understand it correctly? If yes, is there any possibility to "refresh" the tendrl-node-agent state in case it is firstly installed and then the Gluster/Ceph storage cluster is created?
> (14:13:46) rkanade: @dahorak @fbalak I have identified the job routing regression, you guys dont have to do anything, I am sending a PR, please test after that is merged
> (14:13:48) rkanade: ill update here
> (14:14:47) dahorak: rkanade: ok, so it is perfectly ok to install and configure tendrl-node-agents and then install and create Ceph/Gluster cluster? (just to be sure, if we don't need to change our automated playbooks)
> (14:15:48) rkanade: @dahorak I would suggest for import cluster, install and start the sds correctly on all nodes, ensure gluster all nodes are peer probed and only then start tendrl-node-agents on all nodes
> (14:16:50) rkanade: We calculate gluster same $nvr uniqueness based on "gluster pool list" which can change if another node is peer probed after initial cluster detection
> (14:17:33) dahorak: rkanade: ok, so it will be better to start the tendrl-node-agent first time when the gluster or ceph cluster is fully ready, right?
> (14:18:00) rkanade: @dahorak For ImportCluster, Yes
> (14:18:01) dahorak: does this apply somehow to ceph cluster?
> (14:18:08) rkanade: doesnt apply to ceph really
> (14:18:27) dahorak: ok, great...
> (14:18:31) rkanade: thanks

PR #105, opened 2017-05-30
add and use gluster_peers playbook

dahorak:
* in ci_default_import playbook only create gluster trusted storage
pool, do not create bricks (pre-created bricks are not recognized by
Tendrl).


mbukatov:
@dahorak Do we have an issue for that?

dahorak:
Yes, exactly.

PR #107, opened 2017-05-30
add manually user tendrl-user

fbalak:
Add user tendrl-user
WORKAROUND for https://bugzilla.redhat.com/show_bug.cgi?id=1456552#c5
TODO: delete after resolving

### DIFF
--- a/ci_default.yml
+++ b/ci_default.yml
@@ -38,3 +38,13 @@
   tasks:
     - name: Run partprobe
       shell: partprobe || true
+
+# WORKAROUND for https://bugzilla.redhat.com/show_bug.cgi?id=1456552#c5
+# TODO: delete after resolving
+- hosts: all
+  remote_user: root
+  tasks:
+    - name: Add user tendrl-user
+      user:
+        name: tendrl-user
+        group: tendrl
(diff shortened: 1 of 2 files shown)


mkudlej:
There is also issue for this https://github.com/Tendrl/usmqe-setup/issues/108

ci_default.yml
```diff
@@ -38,3 +38,13 @@
   tasks:
     - name: Run partprobe
       shell: partprobe || true
+
+# WORKAROUND for https://bugzilla.redhat.com/show_bug.cgi?id=1456552#c5
+# TODO: delete after resolving
+- hosts: all
```
dahorak: I would suggest to use
`hosts: "usm_server:usm_nodes"`
Instead of `all`, because there might be situation, when inventory file contains additional hosts (for example in CentOS CI it contains also `qe_server`).

ci_default_import.yml
```diff
@@ -36,3 +36,13 @@
   tasks:
     - name: Run partprobe
       shell: partprobe || true
+
+# WORKAROUND for https://bugzilla.redhat.com/show_bug.cgi?id=1456552#c5
+# TODO: delete after resolving
+- hosts: all
```
dahorak: Same as above, `"usm_server:usm_nodes"` instead of `all` would be better.

ci_default.yml
fbalak: Good idea. Changed.

PR #109, opened 2017-05-30
Custom tendrl repo

dahorak:
* rename tendrl-repo role to qe-tendrl-repo
* parametrize qe-tendrl-repo


roles/qe-tendrl-repo/README.rst
```diff
@@ -0,0 +1,19 @@
...
+
+By default `tendrl/release copr`_ and `tendrl/dependencies copr`_ repositories.
+
+The default repositories might be overwritten by following variables:
+  - tendrl_repo_url
+  - tendrl_repo_gpgkey_url
+  - tendrl_dependencies_repo_url
+  - tendrl_dependencies_repo_gpgkey_url
+
+If gpgkey url set to empty string (""), gpgcheck is disabled.
```
mbukatov: Would it make sense to check if the variable is defined?

dahorak: I was thinking about it, but I don't think so, as it is defined in _defaults_. Or is there possibility, that  variable configured in _defaults_ will be undefined?

mbukatov: Ok, that makes sense.

PR #111, opened 2017-06-13
use roles from tendrl-ansible instead of bundled ones

mbukatov:
Remove old usmqe `tendrl-*` roles and use roles from `tendrl-ansible` when possible instead.

This is work in progress, but it seems to work good enough. I'm still going to:

* update documentation/readme files to reflect this new statte
* fix issues (if anything new is discovered)

Resolves https://github.com/Tendrl/usmqe-setup/issues/75


mbukatov:
@fbalak These [roles will be reused from tendrl-ansible](https://github.com/Tendrl/tendrl-ansible/tree/master/roles). I still need to update the docs and readme files to make this obvious, but the idea here is that from now on, we will run the playbook like this:

```
ANSIBLE_ROLES_PATH=/home/usmqe/tendrl-ansible/roles ansible-playbook -i mbukatov-usm1.hosts ci_default.yml
```

mbukatov:
Note: it would not work as nicely as I originally hoped because of https://github.com/Tendrl/tendrl-ansible/issues/10 - but this is not directly caused by changes done here in usmqe-setup.

dahorak:
I'll merge this synchronized with related changes in other repos/places.

PR #120, opened 2017-08-07
add tendrl_unit_test_setup.yml

fbalak:
Add playbook that sets machine up for unit testing.

### DIFF
--- a/tendrl_unit_test_setup.yml
+++ b/tendrl_unit_test_setup.yml
@@ -19,6 +19,11 @@
     - name: Install pytest
       yum:
         name: 'pytest'
+        state: latest
+    - name: Install python-mock
+      yum:
+        name: 'python-mock'
+        state: latest
     - name: Install tendrl packages
       yum:
         name: 'tendrl*'
--- a/tendrl_unit_test_setup.yml
+++ b/tendrl_unit_test_setup.yml
@@ -4,10 +4,6 @@
 - hosts: usm_test
   remote_user: root
   pre_tasks:
-    - name: upgrade all packages
-      yum:
-        name: '*'
-        state: latest
     - name: WORKAROUND configure selinux to permissive mode
       selinux:
         state: permissive
@@ -16,8 +12,12 @@
     - { role: epel, epel_enabled: 1 }
     - qe-tendrl-repo
   post_tasks:
+    - name: upgrade all packages
+      yum:
+        name: '*'
+        state: latest
     - name: Install pytest
-      pip:
+      yum:
         name: 'pytest'
     - name: Install tendrl packages
       yum:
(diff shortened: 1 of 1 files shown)


tendrl_unit_test_setup.yml
```diff
@@ -0,0 +1,35 @@
...
+      retries: 5
+      delay: 5
+    - name: WORKAROUND configure selinux to permissive mode
+      selinux:
+        state: permissive
+        policy: targeted
+  roles:
+    - { role: epel, epel_enabled: 1 }
+    - qe-tendrl-repo
+  post_tasks:
```
mbukatov: You should also make sure `pytest` is installed.

tendrl_unit_test_setup.yml
```diff
@@ -0,0 +1,35 @@
...
+# * Install Tendrl server.
+# * Install Tendrl storage nodes.
+# * Install Gluster nodes.
+# * Configure gluster trusted storage pool.
+# * Install Ceph nodes.
+# * Preconfigure ceph-ansible.
+- hosts: usm_test
+  remote_user: root
+  vars:
+    etcd_ip_address: "{{ ansible_default_ipv4.address }}"
```
mbukatov: This variable doesn't need to be there.

tendrl_unit_test_setup.yml
```diff
@@ -0,0 +1,35 @@
+---
+# Perform following tasks, when applicable
+# (e.g. there are nodes in gluster or ceph_* group)
+# * Configure firewall
+# * Install Tendrl server.
+# * Install Tendrl storage nodes.
+# * Install Gluster nodes.
+# * Configure gluster trusted storage pool.
+# * Install Ceph nodes.
+# * Preconfigure ceph-ansible.
```
mbukatov: This comment seems out of touch with this playbook file :)

tendrl_unit_test_setup.yml
```diff
@@ -29,6 +16,9 @@
     - { role: epel, epel_enabled: 1 }
     - qe-tendrl-repo
   post_tasks:
+    - name: Install pytest
+      pip:
+        name: 'pytest'
```
mbukatov: Could you install pytest from rpm package instead?

tendrl_unit_test_setup.yml
```diff
@@ -0,0 +1,25 @@
+---
+# Sets up machine so there can be performed Tendrl unit tests
+# Installs all tendrl packages and pytest
+- hosts: usm_test
```
dahorak: I'm thinking about better/more descriptive name for the `usm_test` group...  what about something like `unittest_node`?

mbukatov: That is a good point.

fbalak: agreed

PR #126, opened 2017-08-24
Fix gluster repository setup and package installation

mbukatov:
This pull request fixes a rotten gluster setup roles by:

* specifying particular glusterfs releases we can use with the role
* fix wrong way to handle the fact that Storage SIG packages are unsigned
* enable nightly glusterfs builds by default (as suggested by tendrl devs)
* linking to maintainers of the gluster repositories
* removing unnecessary packages during glusterfs installation

### DIFF
--- a/gluster.yml
+++ b/gluster.yml
@@ -5,7 +5,8 @@
 - hosts: gluster
   remote_user: root
   vars:
-    install_from: packages
+    gluster_centos_repo: "nightly"
+    # gluster_centos_repo: "3-11" # for GlusterFS 3.11
   roles:
     - gluster-centos-repo
     - gluster-gdeploy-copr
--- a/gluster.yml
+++ b/gluster.yml
@@ -9,5 +9,5 @@
     # gluster_centos_repo: "3-11" # for GlusterFS 3.11
   roles:
     - gluster-centos-repo
-    - gluster-gdeploy-copr
+    # - gluster-gdeploy-copr
     - gluster-centos
--- a/roles/gluster-centos-repo/README.rst
+++ b/roles/gluster-centos-repo/README.rst
@@ -2,6 +2,30 @@
 Gluster-centos-repo
 ===================
 
-This role adds Gluster repo maintained by `CentOS Storage SIG`_.
+This role adds Gluster repo maintained by `CentOS Storage SIG`_. Note that
+packages installed by this role are not signed.
+
+The role is configured by variable ``gluster_centos_repo`` which could be
+eiter:
+
+* ``nightly``: Latest nightly builds from ``artifacts.ci.centos.org``.
+  When used, ``epel`` role is added into dependencies of this role so that
+  EPEL_ is installed and enabled as well (see explnanation below).
+* ``3-12``: GlusterFS version 3.12
+* ``3-11``: GlusterFS version 3.11
+* ``3-10``: GlusterFS version 3.10
+* ``3-9``: GlusterFS version 3.9
+
+Also note:
+
+* Nightly builds repo is the default.
+* You can't specify repo url directly, you need to select one release from the
+  list shown above.
+* It could happen that for nightly or rc builds, you would need to enable EPEL_
+  via ``epel`` role (eg. when you see error like ``Error: Package ...
+  Requires: liburcu-cds.so.6()(64bit)``, it means you are missing
+  `userspace-rcu`_).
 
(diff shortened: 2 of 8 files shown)


mbukatov:
@fbalak I'm already using these changes when setting up gluster machines, but we can wait for @dahorak to review on Monday.

mbukatov:
Here is an example of storage machine configured by `gluster.yml` playbook using changes in this pull reuqest:

```
[root@mbukatov-usm1-gl2 ~]# yum repolist
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: ftp.cvut.cz
 * epel: download.eng.brq.redhat.com
 * extras: ftp.cvut.cz
 * updates: ftp.cvut.cz
repo id                                                      repo name                                                                                  status
base/7/x86_64                                                CentOS-7 - Base                                                                             9,363
*epel/x86_64                                                 Extra Packages for Enterprise Linux 7 - x86_64                                             11,813
extras/7/x86_64                                              CentOS-7 - Extras                                                                             450
gluster-centos                                               Gluster nightly builds from CentOS Storage SIG                                                 18
updates/7/x86_64                                             CentOS-7 - Updates                                                                          2,146
repolist: 23,790
[root@mbukatov-usm1-gl2 ~]# rpm -qa | grep gluster
glusterfs-libs-4.0dev-0.81.gitee3d9ac.el7.centos.x86_64
glusterfs-client-xlators-4.0dev-0.81.gitee3d9ac.el7.centos.x86_64
glusterfs-cli-4.0dev-0.81.gitee3d9ac.el7.centos.x86_64
python2-gluster-4.0dev-0.81.gitee3d9ac.el7.centos.x86_64
glusterfs-4.0dev-0.81.gitee3d9ac.el7.centos.x86_64
glusterfs-api-4.0dev-0.81.gitee3d9ac.el7.centos.x86_64
glusterfs-fuse-4.0dev-0.81.gitee3d9ac.el7.centos.x86_64
glusterfs-server-4.0dev-0.81.gitee3d9ac.el7.centos.x86_64
glusterfs-geo-replication-4.0dev-0.81.gitee3d9ac.el7.centos.x86_64
```

gluster.yml
```diff
@@ -5,8 +5,9 @@
 - hosts: gluster
   remote_user: root
   vars:
-    install_from: packages
+    gluster_centos_repo: "nightly"
+    # gluster_centos_repo: "3-11" # for GlusterFS 3.11
   roles:
     - gluster-centos-repo
-    - gluster-gdeploy-copr
+    # - gluster-gdeploy-copr
```
fbalak: If you don't want to use this role then it would be better to remove it instead of commenting it out.

roles/gluster-centos-repo/vars/main.yml
```diff
@@ -0,0 +1,17 @@
...
+    3-12:
+        description: "Gluster 3.12 from CentOS Storage SIG"
+        baseurl: "http://buildlogs.centos.org/centos/7/storage/x86_64/gluster-3.12/"
+    3-11:
+        description: "Gluster 3.11 from CentOS Storage SIG"
+        baseurl: "http://buildlogs.centos.org/centos/7/storage/x86_64/gluster-3.11/"
+    3-10:
+        description: "Gluster 3.10 from CentOS Storage SIG"
+        baseurl: "http://buildlogs.centos.org/centos/7/storage/x86_64/gluster-3.10/"
+    3-9:
```
fbalak: We probably don't need this many repositories and we should stick with nightly, 3-12 and maybe 3-11 only.

mbukatov: That's true, we need to verify which versions should be available for testing in upstream and remove the others ...

gluster.yml
mbukatov: I didn't remove it here to highlight that one can still use this role here (in a scenario which is no longer a primary one, but it should do something in upstream).

roles/gluster-centos-repo/README.rst
```diff
@@ -2,6 +2,30 @@
...
+
+Also note:
+
+* Nightly builds repo is the default.
+* You can't specify repo url directly, you need to select one release from the
+  list shown above.
+* It could happen that for nightly or rc builds, you would need to enable EPEL_
+  via ``epel`` role (eg. when you see error like ``Error: Package ...
+  Requires: liburcu-cds.so.6()(64bit)``, it means you are missing
+  `userspace-rcu`_).
```
mbukatov: @mkudlej here it is ...

PR #137, opened 2017-09-18
rpm validation setup: update for 1.5.2

dahorak:
* tendrl_unit_test_setup: enable gluster repo because of gluster dependencies for tendrl-gluster-integration
* do not install tendrl-node-selinux because of conflict with tendrl-server-selinux

### DIFF
--- a/tendrl_unit_test_setup.yml
+++ b/tendrl_unit_test_setup.yml
@@ -11,6 +11,7 @@
   roles:
     - { role: epel, epel_enabled: 1 }
     - qe-tendrl-repo
+    - gluster-centos-repo
   post_tasks:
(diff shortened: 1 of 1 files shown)


tendrl_unit_test_setup.yml
```diff
@@ -28,3 +29,4 @@
       yum:
         name: 'tendrl*'
         state: latest
+        exclude: 'tendrl-node-selinux'
```
mkudlej: Are you sure that this is correct? Is there any issue related to removal of this package?

mkudlej: Ok, my mistake.

PR #140, opened 2017-10-03
Use tendrl-ansible roles from package

dahorak:
add `qe_get_tendrl_ansible.yml` playbook:
- get tendrl-ansible package and unpack it to local directory
  (by default `./tendrl-ansible-tmp/`)
- add the default *tendrl_ansible_dir* (`./tendrl-ansible-tmp/`) to
  `.gitignore`

rename used tendrl-ansible roles:
- roles from tendrl-ansible package are prefixed by `tendrl-ansible.`

### DIFF
--- a/qe_get_tendrl_ansible.yml
+++ b/qe_get_tendrl_ansible.yml
@@ -1,0 +1,49 @@
+---
+
+#
+# Get content of tendrl-ansible package
+#
+
+- hosts: usm_server
+  remote_user: root
+  vars:
+    tendrl_ansible_dir: "./tendrl-ansible-tmp/"
+  roles:
+    - qe-tendrl-repo
+  tasks:
+    # This task is not necessary, but we will at least check that
+    # installation works as expected
+    - name: Install tendrl-ansible package
+      yum: name=tendrl-ansible state=latest
+
+    # prepare local directory for tendrl-ansible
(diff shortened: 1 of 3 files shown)


qe_get_tendrl_ansible.yml
```diff
@@ -0,0 +1,49 @@
...
+  remote_user: root
+  vars:
+    tendrl_ansible_dir: "./tendrl-ansible-tmp/"
+  roles:
+    - qe-tendrl-repo
+  tasks:
+    # This task is not necessary, but we will at least check that
+    # installation works as expected
+    - name: Install tendrl-ansible package
+      yum: name=tendrl-ansible state=latest
```
mbukatov: It would be better to have a separate test case for this. One part of the installation is already covered in yum install rpm test cases, but I agree that there definitely is a room for more testing. I would suggest to draft a test case with some *actual checks* instead of this.

mbukatov: I created an issue for that here: https://github.com/usmqe/usmqe-testdoc/issues/13

dahorak: My intention with installing the tendrl-ansible package on the Tendrl server is mainly to *have it installed* somewhere - for any kind of manual tests/checks/work... (It shouldn't break anything and as a side effect it check, if the installation works.)
I totally agree, that we should test it separately and more deeply.

PR #141, opened 2017-10-05
generate/obtain ssl certificates

dahorak:
- add role for generating/obtaining ssl certificate signed by OOO CA
- generate ssl certs for tendrl


roles/ssl-cert/files/ssl_cert.sh
```diff
@@ -0,0 +1,66 @@
...
+function mrg_ssl_create_key() {
+  local CERT_NAME=$1
+  local CERT_CN=$2
+  test -f /etc/cron.d/redhat-ddns && \
+    local DDNS_NAME=$(head -1 /etc/cron.d/redhat-ddns | sed 's|.*name=\([-a-zA-Z0-9]*\)&.*|\1|').usersys.redhat.com
+  host ${CERT_CN} && local IP=$(host ${CERT_CN} | sed 's|[-0-9a-zA-Z.]* has address \([0-9.]*\)|\1|')
+
+
+  # Prepare openssl.cnf
+  :> /tmp/openssl.cnf
```
mkudlej: please use ```mktemp``` with **RW** for user only

dahorak: @mkudlej addressed by 76a49f129e0f19517eb9aa9f597237b6fa996555

dahorak: The `openssl.cnf` file is in /tmp/ directory, because it might be worth to check how the certificate was configured. @mkudlej do you think, that I should create also the config file in the working (previously created temp directory), so it will be cleaned up together with the tmp dir?

dahorak:
Rebased because of small fixes in the  initial commit.

roles/ssl-cert/tasks/main.yml
```diff
@@ -0,0 +1,51 @@
...
+    mode: 0644
+    owner: "{{ ssl_owner}}"
+    group: "{{ ssl_group }}"
+
+- name: Copy generated key to desired location
+  copy:
+    src: "{{ ssl_cert_tmp_dir.path }}/{{ ssl_cert_name }}.key"
+    dest: "{{ ssl_keys_dir }}/{{ ssl_cert_name }}.key"
+    remote_src: yes
+    mode: 0600
```
mbukatov: It may turn out that I may need to tweak this as well, but let's wait for verification from my side ...

roles/ssl-cert/files/ssl_cert.sh
mkudlej: Well, if you think that you need to check this file because of debug in future, let it like it is.

roles/ssl-cert/tasks/main.yml
mbukatov: Ok, could you allow this to be configured as well? It turns out I would need to set it as `0644` anyway ...

dahorak: @mbukatov  could you try it (with fc0201c6fcf22551d147a92e443436f2fa9c8bda) ... hopefully it will work :)

roles/qe-ssl-cert/files/ssl_cert.sh
```diff
@@ -0,0 +1,66 @@
...
+  echo "emailAddress_max = 40" >> /tmp/openssl.cnf
+  echo "" >> /tmp/openssl.cnf
+  echo "[v3_ca]" >> /tmp/openssl.cnf
+  echo "basicConstraints = CA:FALSE" >> /tmp/openssl.cnf
+  echo "keyUsage = nonRepudiation, digitalSignature, keyEncipherment" >> /tmp/openssl.cnf
+  echo "subjectAltName = @alt_names" >> /tmp/openssl.cnf
+  echo "" >> /tmp/openssl.cnf
+  echo "[alt_names]" >> /tmp/openssl.cnf
+  echo "DNS.1 = ${CERT_CN}" >> /tmp/openssl.cnf
+  echo "DNS.2 = $(hostname -f)" >> /tmp/openssl.cnf
```
mbukatov: Could we check if dns.1 == dns.2 and add the 2nd dns name only if needed?

dahorak: @mbukatov good point! Addressed by a46639b.

PR #142, opened 2017-10-05
Add gdeploy confs

mkudlej:


gdeploy_config/volume_usmqe_beta_arbiter_2_plus_1x2.create.conf
```diff
@@ -0,0 +1,30 @@
...
+
+[volume]
+volname=volume_usmqe_beta_arbiter_2_plus_1x2
+action=create
+brick_dirs=/mnt/brick_usmqe_beta_arbiter_1/1,/mnt/brick_usmqe_beta_arbiter_2/2,/mnt/brick_usmqe_beta_arbiter_3/3
+transport=tcp
+replica=yes
+replica_count=2
+arbiter_count=1
+force=yes
```
dahorak: I would remove the ``force=yes``, because it might cause unexpected behavior without any notification (e.g. RHBZ1498485).

gdeploy_config/volume_usmqe_gama_disperse_4_plus_2x2.create.conf
```diff
@@ -0,0 +1,24 @@
+#   plus 1 block device as arbiter named (/dev/vdd)
```
mbukatov: This comment should be updated to list devices and machines required in the same way as in other gdeploy files.

dahorak: Could you add slightly more descriptive comment here (for me it seems, like there is something missing)?
And maybe it might be worth to update comments for all configs with some description/link what it configure, not only state the requirements.

dahorak: It might be helpful to have link to some description of the particular volume type, but if you think it's ok as it is now, I'm ok with that too.

mkudlej: I think this is good idea, I've added links to US docs.

PR #144, opened 2017-10-12
add configure_email_notifications.yml

dahorak:
- add playbook for configuring email notification for admin user


dahorak:
@mbukatov this playbook is intended just for configuring admin user notification - it expects that Tendrl notifier is already properly configured. It is used in Jenkins job after *tendrl.yml* playbook and only if other variables for tendrl notifier are configured previously. So I wouldn't add the *tendrl-server* role here again, but it might make sense to rename the playbook to be more precise what it do.... something like ``configure_admin_email_notifications.yml``.

mbukatov:
@dahorak could you write down the expectations into the beginning of the file then?

PR #146, opened 2017-10-16
changes for 1.5.3

dahorak:
tendrl.yml: configure etcd_fqdn and graphite_fqdn

### DIFF
--- a/tendrl.yml
+++ b/tendrl.yml
@@ -7,8 +7,9 @@
 - hosts: usm_server
   remote_user: root
   vars:
+    etcd_fqdn: "{{ ansible_fqdn }}"
     etcd_ip_address: "{{ ansible_default_ipv4.address }}"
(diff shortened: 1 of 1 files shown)


tendrl.yml
```diff
@@ -7,8 +7,9 @@
 - hosts: usm_server
   remote_user: root
   vars:
+    etcd_fqdn: "{{ ansible_fqdn }}"
     etcd_ip_address: "{{ ansible_default_ipv4.address }}"
```
mbukatov: With this particular setup, we just need to make sure that `lookup('dig', etcd_fqdn) == etcd_ip_address`, as long as we use IPv4 network.

dahorak: @mbukatov do you think, that it will make sense to use  ```etcd_ip_address: "0.0.0.0"```?

PR #149, opened 2017-10-18
add upstream repos only on CentOS

dahorak:
Playbooks tendrl.yml,  qe-tendrl-repo.yml: add condition for adding upstream repos (add upstream repos only on CentOS)
This change will allow usage of of the playbooks on both CentOS and RHEL.


dahorak:
Yep, there was real syntax issue previously.

dahorak:
I'll merge it once test jobs pass.

PR #150, opened 2017-10-24
make possible epel installation on rhel

dahorak:
* allow epel on rhel, when `qe_not_production_testing_server` parameter is defined

### DIFF
--- a/roles/epel/tasks/main.yml
+++ b/roles/epel/tasks/main.yml
@@ -9,6 +9,7 @@
     - epel_enabled
     - ansible_distribution == 'RedHat'
     - qe_not_production_testing_server is not defined
+    - not qe_not_production_testing_server|bool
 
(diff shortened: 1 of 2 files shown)


roles/epel/tasks/main.yml
```diff
@@ -5,7 +5,10 @@
 
 - name: Check if EPEL is not permanently enabled on RHEL
   fail: msg="Do not permanently enable EPEL on RHEL!"
-  when: epel_enabled and ansible_distribution == 'RedHat'
+  when:
+    - epel_enabled
+    - ansible_distribution == 'RedHat'
+    - qe_not_production_testing_server is not defined
```
mbukatov: Shouldn't we define this variable for particular qe playbooks or qe roles in another commit of this pull request then?

PR #151, opened 2017-11-06
cleanup tendrl.yml playbook

dahorak:
remove unrequired variables


dahorak:
@mbukatov If I understand it correctly, `etcd_ip_address` is required only for *Tendrl server*, am I right? So I removed it from sections for Gluster and Ceph.

PR #157, opened 2017-11-14
test setup for smtp alerting

mbukatov:
Test setup and teardown for email alerting test cases.

### DIFF
--- a/test_setup.smtp.yml
+++ b/test_setup.smtp.yml
@@ -48,11 +48,9 @@
         line: "{{ item.line }}"
         state: present
       with_items:
-        - regexp: '^#? *inet_interfaces = \$myhostname, localhost'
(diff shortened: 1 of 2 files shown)


test_setup.smtp.yml
```diff
@@ -0,0 +1,130 @@
...
+# ========================================
+#  QE configuration of SMTP notifications
+# ========================================
+#
+# Overview of the configuration this playbook sets up:
+#
+# * postfix smtp server runs on usm_client machine, listening on public
+#   interface
+# * tendrl notifier is configured to send email via postfix smtp server running
+#   on usm_client directly (so mx entries are required for this to work)
```
dahorak: I think there should be: *...(so **no** MX entries are required for this to work)*.
Also I would use upper case for *MX*.

PR #158, opened 2017-11-16
QE tools update

mbukatov:
I'm adding:

* ``lsof`` package into qe tools playbook
* new playbook for setup of tmux with my config file

Since tmux config file I used on Fedora no longer works on RHEL 7, setting up tmux was more manual procedure than I liked, so I decided to create a playbook for that.


mbukatov:
All right, removed the fedora conf file.

PR #160, opened 2017-11-24
splitbrain setup and teardown playbook

mbukatov:
Playbooks to invoke split brain on 2 way replicated volume. Functionality verified during work on https://usmqe-testdoc.readthedocs.io/en/latest/web/alerting/glusternative_subvolume.html

### DIFF
--- a/test_teardown.splitbrain.yml
+++ b/test_teardown.splitbrain.yml
@@ -1,23 +1,46 @@
 ---
 
-- hosts: gluster_splitbrain_group1
+- hosts: usm_client[0]
   remote_user: root
   tasks:
-    - name: Allow all traffic to the other group
+    - name: Remove splitbrain invoking iptables rules
       iptables:
+        action: insert
         state: absent
         chain: OUTPUT
         destination: "{{ item }}"
         jump: DROP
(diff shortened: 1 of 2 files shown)


test_teardown.splitbrain.yml
```diff
@@ -0,0 +1,46 @@
...
+      mount:
+        path: "{{ volume_mountpoint }}"
+        fstype: glusterfs
+        src: "{{ groups['gluster'][0] }}:{{ volume_name }}"
+        state: absent
+
+    - name: Remove splitbrain mountpoint directory
+      file:
+        path: "{{ volume_mountpoint }}"
+        state: absent
```
dahorak: Just curious if this is needed?

From [ansible doc](http://docs.ansible.com/ansible/latest/mount_module.html):
> Similarly, specifying `absent` will remove the mount point directory.

mbukatov: You are right, it is not needed.

mbukatov: Fixed in 7cf0ddcb4afc264f1c03af80b7b120b5eed81bd2 and tested that it works.

Issue #161, opened 2017-11-28
Add firewall setup for Tendrl


mbukatov:
Add firewall setup (playbook and a role) for Tendrl based on upstream https://github.com/Tendrl/documentation/wiki/Tendrl-firewall-settings

Issue #70, opened 2017-03-29
 Remove Workaround: stop firewalld on Tendrl server and Tendrl nodes


dahorak:
Remove workaround introduced in commit 76e6e4f2bf35175f4e56a0bb932f657e60fb2852 of pull request #73.

PR #164, opened 2017-12-04
add firewall.tendrl.yml playbook

fbalak:
Based on
https://github.com/Tendrl/documentation/wiki/Tendrl-firewall-settings

Resolves https://github.com/usmqe/usmqe-setup/issues/161


firewall.tendrl.yml
```diff
@@ -0,0 +1,68 @@
...
+      zone=public permanent=true state=enabled immediate=true
+  
+  - name: Enable port for GraphiteWeb
+    firewalld:
+      port=10080/tcp
+      zone=public permanent=true state=enabled immediate=true
+  
+  - name: Enable port for Tendrl (with http)
+    firewalld:
+      port=80/tcp
```
dahorak: It would be slightly nicer, to use `service=http` here instead of `port=...`.

firewall.tendrl.yml
```diff
@@ -0,0 +1,68 @@
...
+      zone=public permanent=true state=enabled immediate=true
+    when: not tendrl_uses_https
+  
+  - name: Enable port for Tendrl (with https)
+    firewalld:
+      port=443/tcp
```
dahorak: Same as above, I would suggest `service=https` instead of `port=...`.

fbalak:
@dahorak I based this playbook on our firewall.skyrings.yml playbook where are tasks arranged in a similar way: https://github.com/usmqe/usmqe-setup/blob/master/roles/firewall-skyrings/tasks/main.yml
It seems more clear to me to set it this way with description containing name of service which port is set but I have no problem with your suggestions and I can change it.

dahorak:
Ok, I agree with this implementation (without some loop)... you are right that this will be more clearer.

fbalak:
@dahorak You are right that it does not make any sense to keep both `firewall_tendrl.yml` and `firewall.tendrl.yml` playbooks but I would keep `firewall.tendrl.yml` as the naming corresponds with `firewall.skyrings.yml` playbook which is not going anywhere. I would wait for @mbukatov's response and after that I will change it according to your suggestion.

dahorak:
@fbalak if you want to remove `firewall_tendrl.yml` and replace it by `firewall.tendrl.yml`, then you will  have to consider the content from the old file and add it to the new one. And also change the reference to this files in other playbooks (I'm aware of: `ci_default_pre_setup.yml`).

mbukatov:
@dahorak's suggestion to preserve content of dropped `firewall_tendrl.yml` seems to make sense at first, but when I review it's actual content, I see that only firewalld installation and ceph firewall roles are there. I don't think that either of those is worth preserving at this moment. We don't test with ceph right now and if we want to keep the ceph tasks there, we should have gluster firewall setup there as well. But we don't do that right now.

For this reason, I would rather drop `firewall_tendrl.yml` as @fbalak did, but as @dahorak point's out, we still needs to change the name of firewall playbook in other playbooks.

Besides that, there is `firewall_tendrl_workaround.yml` which should be dropped as well.

dahorak:
@mbukatov yep, that's good point. I would maybe preserve also the installation of `firewalld`...

@fbalak so my suggestion is:
* add installation of `firewalld` to the `firewall.tendrl.yml` playbook,
* remove  `firewall_tendrl.yml` and `firewall_tendrl_workaround.yml` playbooks,
* update `ci_default_pre_setup.yml` (change reference from `firewall_tendrl.yml` to `firewall.tendrl.yml` and remove workaround playbook).

fbalak:
@dahorak I agree with you. I would just remove `firewall_tendrl_workaround.yml` in separate PR as this PR resolves https://github.com/usmqe/usmqe-setup/issues/161 and another would resolve https://github.com/usmqe/usmqe-setup/issues/70

PR #168, opened 2018-01-09
exclude grafana-4.6.* in tendrl_unit_test_setup.yml

dahorak:
because tendrl-monitoring-integration depends on "grafana < 4.5.3"

Fixing issue with preparation of test server for package validation (for example: https://ci.centos.org/view/tendrl-build-release/job/tendrl-pkgval-X-release-prepare/)

### DIFF
--- a/tendrl_unit_test_setup.yml
+++ b/tendrl_unit_test_setup.yml
@@ -20,6 +20,10 @@
       yum:
         name: 'python-mock'
         state: latest
+    - name: "WORKAROUND - exclude=grafana-4.6.*"
+      lineinfile:
+        path: '/etc/yum.conf'
+        line: 'exclude=grafana-4.6.*'
     - name: Install tendrl packages
       yum:
         name: 'tendrl*'


tendrl_unit_test_setup.yml
```diff
@@ -20,6 +20,10 @@
       yum:
         name: 'python-mock'
         state: latest
+    - name: "WORKAROUND - exclude=grafana-4.6.*"
+      lineinfile:
+        path: '/etc/yum.conf'
+        line: 'exclude=grafana-4.6.*'
```
fbalak: Wouldn't it be better to preserve content of `exclude` variable so it would not break any usage with other playbooks that also set this variable (maybe for future work, we don't have anything like that now)? Maybe something like:
```
  lineinfile:
    path: '/etc/yum.conf'
    regexp: '^(exclude=.*)'
    line: '\1 grafana-4.6.*'
    backrefs=True
```

dahorak: @fbalak Good point, let me try if it will do what we expect...

dahorak: It seems to be not so easy to accomplish this request, so I would keep it in the original shape, as it shouldn't break anything in this particular scenario (it is used on each time freshly installed machine). And still I'm considering it just as a Workaround.

fbalak: Ok, then I approve it.

PR #172, opened 2018-01-16
Enhance ansible2gdeploy

dahorak:
- add possibility to edit multiple files at once
- allow change storage devices (`-s/--storage-devices parameter`)

### DIFF
--- a/bin/ansible2gdeploy.py
+++ b/bin/ansible2gdeploy.py
@@ -71,6 +71,11 @@ def main():
     # open gdeploy config files via plain config parser
     gdeploy_confs = {}
     for gdeploy_conf_file in args.gdeployconf:
+        if not os.path.exists(gdeploy_conf_file):
+            msg = "Specified gdeploy config file '{}' doesn't exists.".format(
+                gdeploy_conf_file)
+            print(msg, file=sys.stderr)
+            return 1
         gdeploy_confs[gdeploy_conf_file] = ConfigParser(allow_no_value=True)
         gdeploy_confs[gdeploy_conf_file].read(gdeploy_conf_file)
 
--- a/bin/ansible2gdeploy.py
+++ b/bin/ansible2gdeploy.py
@@ -22,8 +22,10 @@ import sys
 GLUSTER_SERVER_GROUP = "gluster"
 GLUSTER_CLIENT_GROUP = "usm_client"
 
-
-def main():
+def parse_args():
+    """
+    Parse command line arguments.
+    """
     ap = argparse.ArgumentParser(
         description="update of hosts and devices in gdeploy conf file")
     ap.add_argument(
@@ -47,7 +49,18 @@ def main():
         default="",
         help="output file prefix (added before the name of the input gdeploy conf file) " \
             "If empty, input file is overwritten (in place edit).")
-    args = ap.parse_args()
+    ap.add_argument(
+        "-s",
+        "--storage-devices",
+        dest="storage_devices",
+        help="Coma separated list of available devices for bricks.")
+    return ap.parse_args()
+
+def main():
+    """
+    main function
+    """
+    args = parse_args()
 
     # open gdeploy config files via plain config parser
     gdeploy_confs = {}
@@ -76,8 +89,15 @@ def main():
         clients = inventory.options(GLUSTER_CLIENT_GROUP)
     else:
         clients = []
+
+    # prepare list of storage devices
+    storage_devices = []
+    if args.storage_devices:
+        storage_devices = args.storage_devices.split(",")
+
     print("servers: " + ", ".join(servers), file=sys.stderr)
     print("clients: " + ", ".join(clients), file=sys.stderr)
+    print("devices: " + ", ".join(storage_devices), file=sys.stderr)
 
     # update gdeploy config files
     for gdeploy_conf_file in gdeploy_confs:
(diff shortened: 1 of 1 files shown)


bin/ansible2gdeploy.py
```diff
@@ -24,36 +25,43 @@
...
     ap = argparse.ArgumentParser(
-        description="in place update of hosts in gdeploy conf file")
+        description="update of hosts and devices in gdeploy conf file")
     ap.add_argument(
         "-d",
         "--dry-run",
         action="store_true",
-        help="print to stdout instead of in place edit")
-    ap.add_argument("gdeployconf", help="gdeploy config file to edit")
+        help="print to stdout instead of save to file")
```
mbukatov: We no longer do in place edit?

bin/ansible2gdeploy.py
```diff
@@ -24,36 +25,43 @@
...
+        help="gdeploy config files to edit")
     ap.add_argument(
         "-i",
         dest="inventory",
         action="store",
         help="ansible inventory (aka hosts) file")
     ap.add_argument(
-        "-o",
-        dest="output",
+        "-p",
```
mbukatov: Why do we change command line API? Wouldn't that require update of jobs using this script?

bin/ansible2gdeploy.py
dahorak: You can use *in place* edit - by omitting `-p` (output file prefix) argument.

bin/ansible2gdeploy.py
dahorak: Because parameter `-o` is suitable only for one input/output file, so when I changed the script to handle multiple files at once, I have to somehow handle the possibility to update multiple files at once and do not change the original files. The current usage in CI jobs was to save the updated files with prefix *UPDATED_* (handled by loop in bash), so this change is quite suitable for CI usage. It will require just small update in CI jobs.

dahorak:
Added small enhancement: check if specified gdeploy config file exists (otherwise it will create invalid gdeploy config file containing only section `[hosts]`).

bin/ansible2gdeploy.py
mbukatov: ok

bin/ansible2gdeploy.py
```diff
@@ -64,33 +96,61 @@ def main():
...
-        gdeploy_conf.write(sys.stdout)
-    else:
-        if args.output is not None:
-            output_filename = args.output
+    print("servers: " + ", ".join(servers), file=sys.stderr)
+    print("clients: " + ", ".join(clients), file=sys.stderr)
+    print("devices: " + ", ".join(storage_devices), file=sys.stderr)
+
+    # update gdeploy config files
+    for gdeploy_conf_file in gdeploy_confs:
```
mbukatov: It would be better to iterate on both key and values of `gdeploy_confs` dictrionary - that simplifies the code in the for loop a bit:

```
for gdeploy_filename, gdeploy_conf in gdeploy_confs.items():
   ...
```

Then you can use `gdeploy_conf` instead of  `gdeploy_confs[gdeploy_conf_file]`

bin/ansible2gdeploy.py
```diff
@@ -64,33 +96,61 @@ def main():
...
+            # number of used devices
+            n_devices = len(gdeploy_confs[gdeploy_conf_file].get("backend-setup", "devices").split(","))
+
+            if n_devices > len(storage_devices):
+                msg = "Not enough storage devices for {} - available: {} ({}), required: {}".format(
+                    gdeploy_conf_file, len(storage_devices), ",".join(storage_devices), n_devices)
+                print(msg, file=sys.stderr)
+                return 1
+            gdeploy_confs[gdeploy_conf_file].set("backend-setup", "devices", \
+                ",".join(storage_devices[:n_devices]))
```
mbukatov: I would add here some logging message (in the same way as we do for `print("servers: " ...`) about device allocation, something like:
```
gdeploy_config/volume_gama_disperse_4_plus_2x2.create.conf was assigned devices: sdc,sdd
```

so that when you run it with multiple gdeploy config files, and the 1st one uses them all, the error would be more clear than:

```
devices: sda, sdb, sdc, sdd, sdf
Not enough storage devices for gdeploy_config/volume_gama_disperse_4_plus_2x2.create.conf - available: 0 (), required: 2  
```

bin/ansible2gdeploy.py
```diff
@@ -8,52 +8,84 @@
...
 
 # names of ansible inventory groups to use
 # TODO: make reconfigurable from command line if needed
-GLUSTER_SERVER_GROUP = "gluster"
+GLUSTER_SERVER_GROUP = "gluster-servers"
 GLUSTER_CLIENT_GROUP = "usm_client"
 
-
-def main():
+def parse_args():
```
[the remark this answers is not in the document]
dahorak: The `main` function is already quite long, so this is just "cosmetic" change,... but I can return it back...

bin/ansible2gdeploy.py
dahorak: Good point, fixed by https://github.com/usmqe/usmqe-setup/pull/172/commits/064447e645de405bdeaff333d756b6caa9ea53c0

bin/ansible2gdeploy.py
dahorak: good point, should be fixed by https://github.com/usmqe/usmqe-setup/pull/172/commits/45e77f7e8e517f19676f30f9be6747aee3721e65

PR #173, opened 2018-01-30
[wip] add first version of setup for creating disk with failures

mkudlej:


ltrilety (approved):
looks fine

roles/qe-dmdisks/defaults/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+# Table format, see 'man 8 dmsetup'
+table_format: "flakey"
+# see https://www.kernel.org/doc/Documentation/device-mapper/dm-flakey.txt
+flakey_args: "1 1"
```
fbalak: What arguments are these? `<up interval>` `<down interval>`?

fbalak (commented):
Looks good. Just a few questions...
Also please add PR description.

dahorak:
@mkudlej are the created devices "permanent"? (do they survive reboot?)

roles/qe-dmdisks/defaults/main.yml
mkudlej: Please read document referenced at line 4.

roles/qe-dmdisks/files/gen_dm_file.sh
```diff
@@ -0,0 +1,78 @@
...
+# Get number of sectors till first faulty set of sectors
+block100=$(echo "${first_failure_at}*1024*1024/${sector_size}" | bc)
+
+# Get number of failed sectors in one sector set
+failed_set=$(echo "${failed_block_size}*1024/${sector_size}" | bc)
+
+# Get number of failed sectors till next failure
+repeated_set=$(echo "${repeat_failure_every}*1024*1024/${sector_size}" | bc)
+
+#block100=10
```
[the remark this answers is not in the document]
mkudlej: This code is here so people can try it with human easily countable data sizes. Should I remove it?

mkudlej:
@dahorak I honestly don't know if they are permanent. I think that these changes are not permanent. It is similar as any other device mapper changes made by lvm. You should store them in configuration files, see conf files for lvm disks.

dahorak:
@mkudlej would it be possible to create it permanent (store the configuration in configuration files) directly in the `qe-dmdisks` role or do you think it doesn't make much sense?

dahorak:
@mkudlej I wasn't able to create any volume on the devices :-/ (I've tried it multiple times with similar results).
It seems to me, like it is because of the simulated failures, which prevent even creation of the partitions/bricks and volume. But I'm not sure how to fix it...

roles/qe-dmdisks/defaults/main.yml
mbukatov: ok

mbukatov:
I guess we are trying to hard here, mkfs fails on such device, so there is nothing left to try.

mbukatov:
Actually it would be better to try to:
* stop gluster daemons
* umount brick volume
* setup dm flakey based block device over one for just umounted brick device
* edit fstab to use new flakey device, and mount it again
* start gluster again

Errors simulated by the flakey device should not prevent mounting of the brick and gluster daemon started.

PR #180, opened 2018-03-06
Unify group names with tendrl-ansible

dahorak:
Do we wan't to unify the ansible group names with `tendrl-ansible`?

* use `tendrl-server` instead of `usm_server` ansible group
* use `gluster-servers` instead of `usm_nodes` ansible group

If we decide to merge this PR, it has to be coordinated with changes in other parts of the workflow.

### DIFF
--- a/configure_admin_email_notifications.yml
+++ b/configure_admin_email_notifications.yml
@@ -12,7 +12,7 @@
 #   tendrl_password: password for tendrl user (default: adminuser)
 #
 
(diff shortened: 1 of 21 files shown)


dahorak:
I would vote for not merging this PR at this time - we will use `usm_...` groups in our playbooks from *usmqe-setup* repo and `tendrl-server` and `storage-servers` roles when using *tendrl-ansible* playbooks.

PR #183, opened 2018-03-07
Test setup and teardown for cloud-user like access

mbukatov:
Adds setup and tear down playbook for environment when we don't have direct ssh root access on storage machines and need to connect via different user and then use sudo instead.

### DIFF
--- a/test_setup.clouduser.yml
+++ b/test_setup.clouduser.yml
@@ -9,7 +9,7 @@
 #
 # [1] https://cloudinit.readthedocs.io/en/latest/
 
-- hosts: all
+- hosts: "usm_server:usm_nodes:usm_client"
   remote_user: root
   handlers:
     - name: restart sshd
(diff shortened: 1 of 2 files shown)


mbukatov:
So far, I used this to test changes described in https://github.com/Tendrl/tendrl-ansible/commit/5f53a29588bc7564aa633f880a2bb3d5680bfff7

test_setup.clouduser.yml
```diff
@@ -0,0 +1,81 @@
...
+# account usually created by cloud-init[1] and disables direct root access via
+# ssh.
+#
+# This is usefull for testing ssh access in environment when we don't have
+# direct ssh root access on storage machines and need to connect via different
+# user and then use sudo instead.
+#
+# [1] https://cloudinit.readthedocs.io/en/latest/
+
+- hosts: all
```
dahorak: I wouldn't use `all` here, because this might be quite dangerous for example in CentOS CI, where is also `qe_server` part of the inventory file.
I would propose something like:
`- hosts: "usm_server:usm_nodes:usm_client"`

PR #186, opened 2018-03-20
ansible2gdeploy - host specification type (fqdn, short name, ip, mixed, random)

dahorak:
* add parameter `-h`/`--host-definition {fqdn,short,ip,mixed,random}` (default: `fqdn`)
  Specifies how to define hosts in gdeploy config.

### DIFF
--- a/bin/ansible2gdeploy.py
+++ b/bin/ansible2gdeploy.py
@@ -20,6 +20,8 @@
 from configparser import ConfigParser
 import argparse
 import os
+import random
+import socket
 import sys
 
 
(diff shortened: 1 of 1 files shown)


bin/ansible2gdeploy.py
```diff
@@ -28,6 +32,50 @@
...
+        current_i = 0
+    host_mixed.i = (host_mixed.i + 1) % len(choices)
+    return choices[current_i](fqdn)
+
+def host_random(fqdn):
+    """
+    Return host's fqdn, hostname or IP - randomly selected.
+    """
+    choices = (host_fqdn, host_short, host_ip)
+    return random.choice(choices)(fqdn)
```
mbukatov: This random choice seems nice, but is it really useful when there is no guarantee that each host representation (ip address, short host name and fqdn) is used at least once? That said, I would keep it. I'm just saying I would not use this in automated test setup.

bin/ansible2gdeploy.py
```diff
@@ -104,15 +160,25 @@ def main():
...
     for gdeploy_conf_file, gdeploy_conf in gdeploy_confs.items():
         # ignore hosts section if present
         if gdeploy_conf.has_section("hosts"):
             gdeploy_conf.remove_section("hosts")
         # add servers into hosts sections
         gdeploy_conf.add_section("hosts")
+        print("hosts_definition: %s" % args.hosts_definition, file=sys.stderr)
         for server in servers:
-            gdeploy_conf.set("hosts", server, None)
+            gdeploy_conf.set("hosts", \
```
mbukatov: Nitpicking: why is the `\` at the end of the line?

bin/ansible2gdeploy.py
```diff
@@ -28,6 +32,50 @@
...
+    Return host's fqdn, hostname or IP - sequentially selected.
+    """
+    choices = (host_fqdn, host_short, host_ip)
+    try:
+        current_i = host_mixed.i
+    except AttributeError:
+        host_mixed.i = 0
+        current_i = 0
+    host_mixed.i = (host_mixed.i + 1) % len(choices)
+    return choices[current_i](fqdn)
```
mbukatov: Eh, I don't like this way of doing things, but let's keep it. Could you add here a comment stating something like:

```
# HACK: instead of using a global variable or iterator, we just push an attribute to the function object to preserve global state
```

bin/ansible2gdeploy.py
dahorak: @mbukatov because of pylint complains:
```
C0330: l181: [(bad-continuation), ] Wrong continued indentation (add 9 spaces).
                    host_transformation[args.hosts_definition](server), None)
                    ^        |
```
Adding additional 9 spaces will be definitely more ugly, because the first argument for the method (`"hosts"`) is short, but the second one is much longer.

bin/ansible2gdeploy.py
dahorak: @mbukatov I agree, the `mixed` version is more predictable and definitely better for regular automated test setup. But the `random` function was quite simple, so I put it here... it might be useful sometime.

mbukatov: Agreed.

bin/ansible2gdeploy.py
dahorak: ok, you are right...

PR #191, opened 2018-07-10
setup and teardown for wiki tarball

mbukatov:
This test setup playbook will download enwiki tarball (using url from variable in host inventory file)
and install `wiki-export-split.py` script into `/usr/local/bin` on the client machine.
Teardown just removes both files.

### DIFF
--- a/test_setup.wiki_tarball.yml
+++ b/test_setup.wiki_tarball.yml
@@ -1,0 +1,49 @@
+---
+# This playbooks downloads tarball with a dump of all english wikipedia
+# articles {{ wiki_tarball_filename }} on a client machine, so that
+# one can try to use it for testing, eg:
+#
+#  * extract some articles into individual files on gluster volume via
+#  wiki-export-split.py script, distributing lot of small files on all bricks
+#  so that the files on the bricks will be distributed uniformly
+#
+#  * copy the tarball into the volume directly (this will be a bit faster than
+#  generating equivalent size of random data)
+
+- hosts: usm_client
+  remote_user: root
+  vars:
+    - wiki_tarball_filename: "enwiki-latest-pages-articles.xml.bz2"
+    - wiki_tarball_dest: "/tmp"
+
+  tasks:
+
+    - name: Check that mandatory variables are defined
+      assert:
+        that:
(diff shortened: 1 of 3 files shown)


mbukatov:
I tested this on my machines, the variables I need to set looks like this:

```
wiki_tarball_url="http://ooo.example.com/wikipedia-dump/enwiki-latest-pages-articles.xml.bz2"
wiki_tarball_checksum="md5:a538459bfe1bc35e4d8161840eadbcd0"  
```

test_setup.wiki_tarball.yml
```diff
@@ -0,0 +1,56 @@
...
+          - "You need to define *url of the tarball* in wiki_tarball_url"
+          - "variable in your invenory file."
+          - "Don't use original source, but host the file on your own machine!"
+          - "Eg. http://example.com/enwiki-latest-pages-articles.xml.bz2"
+          - ""
+          - "You also need to define *checksum of the tarball* in"
+          - "wiki_tarball_checksum variable."
+          - "Eg. md5:a538459bfe1bc35e4d8161840eadbcd"
+
+    - name: Download the wiki tarball into /tmp directory
```
fbalak: Name hints that it is downloaded to /tmp but destination is variable. It would be good to rename it to something like: wiki_tarball_dest (default is /tmp).

mbukatov: The point of this task name is to communicate where the tarball will be stored, moreover you are not expected to change the wiki_tarball_dest variable. I will add a variable name into the name though.

test_setup.wiki_tarball.yml
```diff
@@ -0,0 +1,56 @@
...
+  tasks:
+
+    - name: Check that mandatory variables are defined
+      assert:
+        that:
+          - wiki_tarball_url is defined
+          - wiki_tarball_checksum is defined
+        msg:
+          - "You need to define *url of the tarball* in wiki_tarball_url"
+          - "variable in your invenory file."
```
dahorak: s/invenory/inventory/

mbukatov: fixed

PR #195, opened 2018-07-17
Add test setup for unmanage

ebondare:


test_setup.unmanage.yml
```diff
@@ -0,0 +1,81 @@
...
+#
+# The playbook expects inventory file with the following groups:
+#
+# * tendrl-server: a single machine where Tendrl Server (management console) is
+#                  installed
+# * gluster-servers: gluster storage servers (machines with bricks)
+
+
+# Stop individual services on the Storage Nodes
+- hosts: gluster-servers
```
mbukatov: We should use `gluster_servers` group instead. The name was changed this way in tendrl-ansible to prevent problems with future version of ansible.

Issue #197, opened 2018-07-23
Deprecation warning because of result|success


ebondare:
In some roles the syntax "result|success" is used. Ansible 2.6.0 issues a warning that this syntax is deprecated and "result is success" should be used instead.

Here is the list of roles that have that issue: 
```
./qe-server/tasks/main.yml
./rh-python36/tasks/main.yml
./rh-python35/tasks/main.yml
./ceph-centos-repo/tasks/main.yml
./gluster-client/tasks/main.yml
./qe-munin-node/tasks/main.yml
./firewall-gluster/tasks/main.yml
./qe-evidence-probe-journald/tasks/main.yml
./qe-evidence-probe/tasks/main.yml
./epel/tasks/main.yml
./gluster-server/tasks/main.yml
```

mbukatov:
Resovled by https://github.com/usmqe/usmqe-setup/pull/198

PR #198, opened 2018-07-24
Changed syntax, Issue #197

ebondare:
Changed "result|success" to "result is success" due to deprecation warnings from ansible.

### DIFF
--- a/firewall.tendrl.yml
+++ b/firewall.tendrl.yml
@@ -12,7 +12,7 @@
         name: 'firewalld'
         state: present
       register: task_result
-      until: task_result|success
+      until: task_result is success
       retries: 5
       delay: 5
 
--- a/qe_pre_installation_tasks.yml
+++ b/qe_pre_installation_tasks.yml
@@ -20,7 +20,7 @@
       retries: 5
       delay: 5
       register: result
-      until: 'result|success'
+      until: 'result is success'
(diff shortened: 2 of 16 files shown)


firewall.tendrl.yml
```diff
@@ -12,7 +12,7 @@
         name: 'firewalld'
         state: present
       register: task_result
-      until: task_result|success
+      until: task_result is success
```
dahorak: I think, that `irewall.tendrl.yml` file is not part of this repository anymore, so you don't need to change this file to avoid conflict during rebase.

mbukatov: Yep, I will resolve this during merge.

qe_pre_installation_tasks.yml
```diff
@@ -20,7 +20,7 @@
       retries: 5
       delay: 5
       register: result
-      until: 'result|success'
+      until: 'result is success'
```
mbukatov: @dahorak btw do you know why we have this this statement in quotes in some places, but without quotes in others?

dahorak: @mbukatov good question... and honest answer is, that I'm not sure... I think, that it should be safe to use it without quotes, but I'm not sure if there is any rule...

PR #202, opened 2018-09-11
add WORKAROUND: update ansible

dahorak:
- Tendrl/commons/issues/1054


ci_default_post_tendrl.yml
```diff
@@ -17,3 +17,8 @@
 
     - name: Copy usm.ini template
       local_action: template src=./templates/usm.ini.j2 dest=./usm.ini
+
+    - name: "WORKAROUND: update ansible (Tendrl/commons/issues/1054)"
+      yum:
+        name: ansible
+        state: latest
```
mbukatov: This looks innocent, but would it have any side effect on downstream?

PR #204, opened 2018-09-11
Test setup tools: stress-ng

mbukatov:
Adding playbook for installation of tools, which are required for running automated test
cases from `usmqe-tests` repository. So far, only `stress-ng` is included.


mbukatov:
@dahorak do you think that we will need other tools to be installed like stress-ng?

test_teardown.stress_ng.yml
```diff
@@ -0,0 +1,18 @@
...
+  remote_user: root
+  tasks:
+    - name: Remove stress-ng tool and direct dependencies
+      yum:
+        name: "{{ item }}"
+        state: absent
+      with_items:
+        - stress-ng
+        - libbsd
+        - lksctp-tools
```
mbukatov: This `lksctp-tools` is from rhel channel, @dahorak do you think we want to remove this here as well?

dahorak: If we include `lksctp-tools`, we should also consider `libaio`, as this was also installed as dependency of `stress-ng` (from rhel repo) - at least on my setup. But I wouldn't remove them at all, because we are not easily able to distinguish, if they were installed only as dependency of `stress-ng` or as a dependency also for something else and it might remove unintentionally some other packages.

mbukatov: Makes sense, I'm going to remove `lksctp-tools` from there.

test_teardown.stress_ng.yml
```diff
@@ -0,0 +1,17 @@
+---
+# Removal of stress-ng tool and all it's dependencies (checked via yum.log),
```
dahorak: To be slightly more precise - I would mention here, that we are removing only dependencies from EPEL repository (it is partially mentioned on the next line, but still I would slightly rephrase it to be clear).

mbukatov: Good catch, will fix this.

PR #205, opened 2018-09-12
WORKAROUND: restart tendrl-node-agent service

dahorak:
- when *ansible* is updated, restart related *tendrl-node-agent* service
- related to #202

### DIFF
--- a/ci_default_post_tendrl.yml
+++ b/ci_default_post_tendrl.yml
@@ -22,3 +22,10 @@
       yum:
         name: ansible
         state: latest
+      register: ansible_update_status
+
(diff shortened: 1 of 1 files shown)


ci_default_post_tendrl.yml
```diff
@@ -22,3 +22,10 @@
       yum:
         name: ansible
         state: latest
+      register: ansible_update_status
+
+    - name: restart tendrl services
```
mbukatov: Could you add "WORKAROUND" prefix to a task name? Just to make this clear later in ansible log.

dahorak: Good point! I have it updated, but not pushed here...

PR #215, opened 2018-10-11
add gdeploy config for fips mode

mbukatov:
This configures gluster volumes to use `fips-mode-rchecksum`.

Validation:

```
# gluster volume get volume_beta_arbiter_2_plus_1x2 fips-mode-rchecksum
Option                                  Value                                   
------                                  -----                                   
storage.fips-mode-rchecksum             enable
```


fbalak:
Wouldn't be better to have just one playbook to start fips parametrized with variables? This way we have to maintain multiple files.

PR #222, opened 2018-11-26
Alert logger

fbalak:
This playbook creates `alerts_logger_<tendrl-user>` service and enables it on `usm_server` machine.
It periodically collects alerts from tendrl and logs them.


templates/alerts_logger.service.j2
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Tendrl alerts logger for user {{ tendrl_user }}
```
mbukatov: Instead of using jija template and generate actual system unit file, you can use template feature of system uni files, see: https://fedoramagazine.org/systemd-template-unit-files/

That way, you don't need to edit/generate systemd unit file, and you can pass the parameter as a part of service name to start/stop.

test_setup.alerts_logger.yml
```diff
@@ -0,0 +1,36 @@
...
+- hosts: usm_server
+  remote_user: root
+  vars:
+    tendrl_user: "admin"
+    tendrl_password: "adminuser"
+
+  tasks:
+    - name: Create alerts_logger python script based on a template
+      template:
+        dest: /usr/bin/alerts_logger_{{ tendrl_user }}.py
```
mbukatov: I would rather put it into `/usr/loca/bin` (which is better suited for executables which are not installed via rpm).

test_setup.alerts_logger.yml
```diff
@@ -0,0 +1,36 @@
...
+        src: alerts_logger.py.j2
+        force: yes
+        mode: "+x"
+
+    - name: Create service file based on a template
+      template:
+        dest: /lib/systemd/system/alerts_logger_{{ tendrl_user }}.service
```
mbukatov: It would be better to put it into `/etc` instead, as `/lib/systemd` contains the packaged defaults.

templates/alerts_logger.py.j2
```diff
@@ -0,0 +1,144 @@
...
+import os
+import json
+import time
+
+import requests
+
+
+URL = "http://0.0.0.0/api/1.0/"
+USER = "{{ tendrl_user }}"
+PASSWORD = "{{ tendrl_password }}"
```
dahorak: Same as above, wouldn't be better, to read those parameters from command line arguments or some configuration file?

fbalak:
Example of alert format:
```
Dec 10 14:31:47 fbalak-usm2-server.usmqe sh[6916]: INFO:tendrl-notifier.admin:{u'acked_at': u'2018-12-10T09:34:14.706633+00:00', u'delivered': True, u'alert_type': u'UTILIZATION', u'resource': u'memory_utilization', u'severity': u'INFO', u'classification': u'["node"]', u'tags': {u'integration_id': u'2e9892d5-6d0d-4be6-bb57-9cdd98f68bcc', u'cluster_short_name': u'2e9892d5-6d0d-4be6-bb57-9cdd98f68bcc', u'clear_alert': u'CRITICAL', u'fqdn': u'fbalak-usm2-gl2.usmqe', u'warning_max': 90, u'message': u'Memory utilization on node fbalak-usm2-gl2.usmqe in 2e9892d5-6d0d-4be6-bb57-9cdd98f68bcc back to normal'}, u'time_stamp': u'2018-12-10T09:34:14.293024+00:00', u'pid': u'3245', u'acked': True, u'updated_at': u'2018-12-10 09:36:21.986763+00:00', u'ackedby': u'TENDRL', u'source': u'GRAFANA', u'alert_id': u'81824fb6-0b35-4dd1-9e8b-2692c794ff40', u'node_id': u'77bd7650-e630-4229-bffa-f4d591443530', u'ack_comment': u'["System acked"]', u'significance': u'HIGH', u'current_value': u'', u'hash': u'be6a57554aedfc1d434b5fc6d1f739810832653cc68af48a9468fec7fc7891d5a8db0baad997246b6a4969ea9932e3545ebd3c1b1deccc9951e40ce634995b97'}
```

test_setup.alerts_logger.yml
```diff
@@ -0,0 +1,41 @@
...
+      copy:
+        dest: /etc/systemd/system/
+        src: alerts_logger@.service
+
+    - name: force systemd to reread configs
+      systemd: daemon_reload=yes
+
+    - name: Add user to users.ini
+      ini_file:
+        path: /usr/local/bin/users.ini
```
dahorak: I would rather propose some other place for the configuration file (somewhere in `/etc/...`). And also some better (more explanatory) name - something like `usmqe_alerts_logger_users.ini` or similar.

dahorak:
I have one last optional note - wouldn't be better to name the script and service with prefix `usmqe` or something like that - to clearly distinguish, that the service *belongs* to our team?

files/alerts_logger.py
```diff
@@ -0,0 +1,150 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("user", type=str, nargs='?')
+    args = parser.parse_args()
+    config = ConfigParser.RawConfigParser()
+    config.read("/etc/usmqe_alerts_logger_users.ini")
+    password = config.get(args.user, "password")
+    auth = login(args.user, password)
+    api = TendrlApi(auth=auth)
+    logging.basicConfig(level=os.environ.get("LOGLEVEL", "INFO"))
+    logger = logging.getLogger("tendrl-notifier.{}".format(args.user))
```
dahorak: The logs are in fact not from `tendrl-notifier`, but from *usmqe alerts_logger*, so I would reflect it in the logger name. To be easily possible to distinguish/grep logs from core *Tendrl* components and our own logs.

test_setup.alerts_logger.yml
```diff
@@ -0,0 +1,41 @@
+---
+# ========================================
+#  QE configuration for alerts logger
+# ========================================
+#
+# This playbook serves as setup of alerts logger on tendrl server machine.
+# It periodically logs alerts from tendrl api based on tendrl user.
+
+- hosts: usm_server
```
mbukatov: I would rather have this installed on a client machine, as the snmp and smtp alerting logs are stored there, so that we will have all alerting logs on the same machine. Moreover it would keep our usm server pristine.

files/usmqe_alerts_logger.py
```diff
@@ -0,0 +1,150 @@
...
+import ConfigParser
+import logging
+import os
+import json
+import time
+
+import requests
+
+
+URL = "http://0.0.0.0/api/1.0/"
```
mbukatov: Could we have this url loaded from the config file as well? One problem here is that if ssl is enabled, this url may or may not work. The other thing is that I would like to have the script installed on the client machine.

bin/usmqe_alerts_logger.py
```diff
@@ -0,0 +1,140 @@
...
+    password = config.get(args.user, "password")
+    ssl = True if config.get(args.user, "ssl") == "true" else False
+    if ssl:
+        url = "https://" + config.get(args.user, "url") + "/api/1.0/"
+    else:
+        url = "http://" + config.get(args.user, "url") + "/api/1.0/"
+    auth = login(args.user, password, url)
+    api = TendrlApi(auth=auth, url=url)
+    logging.basicConfig(level=os.environ.get("LOGLEVEL", "INFO"))
+    logger = logging.getLogger("usmqe_alerts_logger@{}".format(args.user))
```
mbukatov: There is one problem with this, your new logger reports every log entry and then handles it to the parent logger, so that there are 2 log entries reported for every entry, see: https://docs.python.org/2/library/logging.html#logging.Logger.propagate

The easiest thing (not sure if canonically correct) would be to  disable this propagation by adding line:

```
logger.propagate = False
```

bin/usmqe_alerts_logger.py
```diff
@@ -0,0 +1,140 @@
...
+    formatter = TendrlFormatter(logging.BASIC_FORMAT)
+    handler = logging.StreamHandler()
+    handler.setFormatter(formatter)
+    logger.addHandler(handler)
+    last_alerts = []
+    while(True):
+        alerts = api.alerts()
+        for alert in alerts:
+            if alert not in last_alerts:
+                logger.info(alert)
```
mbukatov: I have additional suggestion, could we also report number of alerts which we don't log again because we just seen them last time? Something like:

```
diff --git a/bin/usmqe_alerts_logger.py b/bin/usmqe_alerts_logger.py
index 6a4fdd3..d770e22 100644
--- a/bin/usmqe_alerts_logger.py
+++ b/bin/usmqe_alerts_logger.py
@@ -129,9 +129,15 @@ def main():
     last_alerts = []
     while(True):
         alerts = api.alerts()
+        # how many alerts in current alert list we have already seen last time
+        seen_count = 0
         for alert in alerts:
             if alert not in last_alerts:
                 logger.info(alert)
+            else:
+                seen_count =+ 1
+        if seen_count > 0:
+            logger.info("not reported %d alerts seen in previous check", seen_count)
         last_alerts = alerts
         time.sleep(10)
```

Would it make sense?

bin/usmqe_alerts_logger.py
```diff
@@ -122,16 +122,25 @@ def main():
...
+        seen_count = 0
         for alert in alerts:
             if alert not in last_alerts:
                 logger.info(alert)
+            else:
+                seen_count =+ 1
+        if seen_count > 0:
+            logger.info("Old alerts seen in last check that were not "\
+                "logged again: {0}".format(
+                    seen_count))
```
mbukatov: That said, we need to use the loggers properly ...

mbukatov: log message should be separate from the value/object dumped used in the message, so that when the log level prevents the log message being passed further, the object doesn't have to be serialized ...

PR #224, opened 2018-11-30
Stop tendrl

fbalak:


dahorak:
I would probably chose some other names for the playbooks (especially for the ` test_teardown.stop_tendrl_nodes.yml`). I understand the naming scheme - but in this case the string *stop_tendrl_nodes* for playbook starting all services seems to be quite confusing (in my opinion).

PR #226, opened 2018-12-05
test setup for ssl

mbukatov:
Work in progress, remaining items:

* [x] variable for disabling port 80
* [x] add redirect setup
* [x] add playbook for setup using QE CA
* [x] resolve comments from others (so far)
* [x] make sure QE CA is installed on qe server and client (for us to be able to receive alerts via https)
* [x] account for problem with custom CA use case

Based on https://github.com/Tendrl/documentation/wiki/Enabling-Https-on-tendrl-server, some code reused from https://github.com/Tendrl/tendrl-ansible/pull/46/files.


roles/tendrl-httpd-ssl/handlers/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+
+- name: restart httpd
+  service:
+    name=httpd
+    state=restarted
```
dahorak: Do we need full restart of `http` service? Wouldn't be enough `state: reloaded`?

roles/tendrl-httpd-ssl/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
+---
+
+- debug:
+    var: httpd_server_name
+
+- debug:
+    var: httpd_ip_address
```
dahorak: Would it make sense, to use some defaults for `httpd_server_name` and for `httpd_ip_address`?

test_setup.tendrl_httpd_ssl_selfsigned.yml
```diff
@@ -0,0 +1,17 @@
...
+#  Setup of HTTPS for Tendrl UI, API and Grafana
+# ===============================================
+#
+# This is **minimal setup**, which **uses default local ssl certs** provided by
+# mod_ssl rpm package.
+
+- hosts: usm_server
+  remote_user: root
+  handlers:
+    - name: restart httpd
```
dahorak: The handler is already defined in the role, isn't it? Or did I miss something?

mbukatov: good point

roles/tendrl-httpd-ssl/handlers/main.yml
mbukatov: good point, fixed: 7bbe9c0c78743a19dd5e2c814947afc5e0cddbcb

roles/tendrl-httpd-ssl/tasks/main.yml
mbukatov: Actually, yes.

mbukatov:
@dahorak the pull request is complete, and includes 2 fixes wrt original steps on the wiki

mbukatov:
@ebondare when the final draft of the ssl docs is available, please check that this pull request matches it

dahorak:
Merging in this state, will create new PR if necessary once documentation for SSL configuration will be finalized.

PR #228, opened 2019-01-04
start and enable atd

fbalak:
Make sure that `atd` is running because of https://github.com/usmqe/usmqe-tests/pull/215.


fbalak:
In that case I would edit `ci_default_post_tendrl.yml`. I don't think that this basic check should be in a separate playbook.

mbukatov:
Ok.

PR #231, opened 2019-01-21
add setup and teardown playbooks for stress tool

fbalak:
This tool will be used for memory utilization as `stress-ng` utlizes cache memory but `stress` utilizes used memory better.


mbukatov:
@fbalak could you also create a pull request which renames the playbook in usmqe-tests?

PR #233, opened 2019-02-18
implement changes in ssl setup

dahorak:
This change is based on:

* https://github.com/Tendrl/api/issues/458 (rename Rename tendrl-ssl.conf.sample)
* https://github.com/Tendrl/api/issues/460 (Change redirect directive, VirtualHost `_default_`)
* ssl setup wikipage: https://github.com/Tendrl/documentation/wiki/Enabling-Https-on-tendrl-server

### DIFF
--- a/roles/tendrl-httpd-ssl/tasks/main.yml
+++ b/roles/tendrl-httpd-ssl/tasks/main.yml
@@ -11,15 +11,15 @@
     name: mod_ssl
     state: latest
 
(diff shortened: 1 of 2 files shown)


dahorak:
> We need to change `test_teardown.tendrl_httpd_ssl.yml` accordingly.
Good point.

dahorak:
Since there will be probably be one other change (name of the sample configuration file should also contain the prefix `00_`, we can wait with the merge of this patch and do the additional update as part of this PR.

PR #245, opened 2019-04-09
test setup: tendrl nodeagent stopped on one node

mbukatov:
These playbooks serve as a setup/teardown for negative test cases, such as import with cluster in an invalid state.


mbukatov:
I may update these playbooks based on the needs of automated cases I'm working on.

